PR #1, opened 2016-07-11
initial checkin based from mbaas-monitoring

jessesarn:


jessesarn:
Closing this as we are going to make separate PR's for the individual JIRA's for each piece so far.  I'll create those as soon as we get the base merged in.

PR #2, opened 2016-07-13
Rhmap 8600 initial nagios3 docker setup

mikenairn:
* Cut down version of https://github.com/feedhenry/nagios3-container/pull/1 with the minimal setup required to run nagios. Additional checks will be added by separate PRs.

Verification:

```
docker build -t docker.io/rhmap/nagios3:0.0.1 .
docker run -i -e NAGIOS_USER='nagiosadmin' -e NAGIOS_PASSWORD='password' -e RHMAP_ROUTER_DNS='rhmap.local.feedhenry.io' -p 8888:8080 -t docker.io/rhmap/nagios3:0.0.1
```

Should start up OK and nagios dashboard accessible at 192.168.33.12:8888


Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM centos:centos7
+
+EXPOSE 8080
+
+ADD sendEmail-epel-7.repo /etc/yum.repos.d/
```
rhcarvalho: Is it fine to use EPEL @vibe13?

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+                                           supervisor \
+                                           python-jinja2 \
+                                           nagios-plugins-all \
+                                           sendEmail \
+                                           perl-Net-SSLeay \
+                                           perl-IO-Socket-SSL && \
+    yum clean all && \
+    mkdir -p /opt/rhmap/ && \
+    sed -i -e 's/Listen 80/Listen 8080/' /etc/httpd/conf/httpd.conf && \
+    sed -i -e 's|DocumentRoot "/var/www/html"|DocumentRoot "/usr/share/nagios/html"|' -e 's|<Directory "/var/www">|<Directory "/usr/share/nagios/html">|' /etc/httpd/conf/httpd.conf && \
```
rhcarvalho: Since the 3 replacements are in the same file, why do we have 1 + 2 replacements in two `sed` invocations?

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM centos:centos7
+
+EXPOSE 8080
+
+ADD sendEmail-epel-7.repo /etc/yum.repos.d/
+
+RUN yum install -y epel-release && \
+    yum -y --setopt=tsflags=nodocs update && \
```
rhcarvalho: Is this necessary? I think it's normally the role of the base image to have up-to-date packages.

I went to look at https://github.com/sclorg/postgresql-container/blob/master/9.5/Dockerfile, and see that we don't run `yum update`.

We do run other things that would be useful here, namely that `rpm -V` to make sure that all packages we passed to `yum install` actually got installed.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+                                           python-jinja2 \
+                                           nagios-plugins-all \
+                                           sendEmail \
+                                           perl-Net-SSLeay \
+                                           perl-IO-Socket-SSL && \
+    yum clean all && \
+    mkdir -p /opt/rhmap/ && \
+    sed -i -e 's/Listen 80/Listen 8080/' /etc/httpd/conf/httpd.conf && \
+    sed -i -e 's|DocumentRoot "/var/www/html"|DocumentRoot "/usr/share/nagios/html"|' -e 's|<Directory "/var/www">|<Directory "/usr/share/nagios/html">|' /etc/httpd/conf/httpd.conf && \
+    touch /supervisord.log supervisord.pid && \
```
rhcarvalho: Maybe it should be absolute path, `/supervisord.pid`?
I see the absolute path is referenced later on.

Since we don't set the WORKDIR, I'd avoid using relative paths.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    mkdir -p /var/log/nagios/rw/ && \
+    mkdir -p /var/log/nagios/spool && \
+    mkdir -p /var/log/nagios/spool/checkresults && \
```
rhcarvalho: `mkdir` can take multiple DIRECTORY arguments.

Also, this line and the above are redundant, because the latter is a subdirectory of the former.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    chmod -R 777 /supervisord.log /supervisord.pid /var/log/nagios \
+                    /etc/httpd /etc/passwd /var/log /etc/nagios /usr/lib64/nagios /var/spool/nagios /run /usr/share/httpd /usr/share/nagios && \
```
rhcarvalho: The indentation here is strange.

Perhaps a comment above the `chmod` explaining why we need to loosen permissions for those particular files/dirs wouldn't do bad.

Not obvious, Dockerfile comments work just fine in the middle of RUN instructions. E.g.: https://github.com/sclorg/mongodb-container/blob/master/3.2/Dockerfile#L33

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    sed -i -e 's|cfg_file=/etc/nagios/objects/localhost.cfg||' /etc/nagios/nagios.cfg
+
+ADD supervisord.conf /etc/supervisord.conf
+ADD make-nagios-fhservices-cfg.py /opt/rhmap/make-nagios-fhservices-cfg.py
+ADD make-nagios-commands-cfg.py /opt/rhmap/make-nagios-commands-cfg.py
+ADD fhservices.cfg.j2 /opt/rhmap/fhservices.cfg.j2
+ADD commands.cfg.j2 /opt/rhmap/commands.cfg.j2
```
rhcarvalho: You can combine the 4 `ADD` instructions above in a single instruction. That makes builds faster and produces less layers. Even if we squash the images later, this is still beneficial -- even the squashing process becomes faster.

https://docs.docker.com/engine/reference/builder/#/add

```Dockerfile
ADD make-nagios-fhservices-cfg.py make-nagios-commands-cfg.py fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/

# Or
ADD *-cfg.py *.cfg.j2 /opt/rhmap/

# Or place the files relative to this Dockerfile in such a way that you can reference all of them in a single word.
ADD rhmap/ /opt/rhmap/

# Or have a "root tree" relative to the Dockerfile and add everything in a single instruction:
# (e.g. https://github.com/sclorg/mongodb-container/blob/master/3.2/Dockerfile#L49)
ADD root/ /
```

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+ADD start.sh start.sh
```
rhcarvalho: Again here I'd avoid using a relative destination, since we don't set WORKDIR.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+CMD ["./start.sh"]
```
rhcarvalho: Please use an absolute path here too.

make-nagios-commands-cfg.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
+import os
+
+from jinja2 import Environment, FileSystemLoader, Template
+
+smtp_server = os.getenv('SMTP_SERVER', 'localhost')
```
rhcarvalho: Shouldn't we fail if the SMTP server configuration is not given, instead of generating dummy config?

start.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/bash
+
+# Add nagios dirs
+mkdir -p /var/log/nagios/archives \
+      /var/log/nagios/rw \
+      /var/log/nagios/spool/checkresults
```
rhcarvalho: Isn't this already created by the Dockerfile?!

start.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Add nagios user
+htpasswd -c -b -s /etc/nagios/passwd ${NAGIOS_USER} ${NAGIOS_PASSWORD}
+
+# Generate command config
+python /opt/rhmap/make-nagios-commands-cfg.py
```
rhcarvalho: The python file should be executable and has a shebang, so no need to call the interpreter here:

```
/opt/rhmap/make-nagios-commands-cfg.py
```

Same below.

start.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/bash
+
```
rhcarvalho: Please add here the safety options:

```
set -o errexit
set -o nounset
set -o pipefail
```

start.sh
```diff
@@ -0,0 +1,17 @@
...
+# Add nagios user
+htpasswd -c -b -s /etc/nagios/passwd ${NAGIOS_USER} ${NAGIOS_PASSWORD}
+
+# Generate command config
+python /opt/rhmap/make-nagios-commands-cfg.py
+
+# Generate fhservices config
+python /opt/rhmap/make-nagios-fhservices-cfg.py
+
+/usr/bin/supervisord
```
rhcarvalho:
```
exec /usr/bin/supervisord
```

I'm not sure `supervisord` works fine with Docker. For sure we should exec it to make it run as PID 1, otherwise the shell won't handle signals properly / won't forward signals to `supervisord`.

Dockerfile
vibe13: I won't be able to use packages from EPEL repositories, so if you need any package from EPEL, please reach me asap :-)

start.sh
mikenairn: Added

Dockerfile
mikenairn: I'd guess it has to run update since we are adding another repo for sendEmail

start.sh
mikenairn: Done

make-nagios-commands-cfg.py
mikenairn: Unsure about this tbh. Being able to send email notifications isn't a mandatory part of nagios even though we do want to have it configurable. Alternative to have it fallback here would be to add default setting to the openshift template, or allow this to work if the smtp settings don't exist.

Dockerfile
mikenairn: No new packages being added (yet).

make-nagios-fhservices-cfg.py
```diff
@@ -0,0 +1,28 @@
...
+nagios_config_filename = '/etc/nagios/conf.d/fhservices.cfg'
+
+template_basename = os.path.basename(template_file)
+template_dirname = os.path.dirname(template_file)
+
+j2env = Environment(loader=FileSystemLoader(template_dirname), trim_blocks=True)
+j2template = j2env.get_template(template_basename)
+
+j2renderedouput = j2template.render(fh_services_ping = fh_services_ping,
+                                    fh_services_health = fh_services_health,
```
rhcarvalho: Minor thing, for consistency with the two lines below (and Python's PEP8), please remove the spaces around the equal sign in the lines above.

Dockerfile
rhcarvalho: `yum install` refreshes the metadata, and you can force it with `--refresh`.
Running `yum update` is not a problem, just something I feel most of the times we leave up to the base image.

If we run a docker build using the cache, it won't "update" packages anyway.

It also creates a layer that is potentially larger than it needs to be: we'd expect this image to be centos + (nagios + related deps), but we end up with centos + (nagios + related deps + updates for unrelated packages).

This also means that when @vibe13 builds the official image, we may have no clue what packages get updated as part of that `yum update`. @vibe13, is this a problem for you?

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    yum -y --setopt=tsflags=nodocs update && \
+    yum -y --setopt=tsflags=nodocs install httpd \
+                                           nagios \
+                                           telnet \
+                                           supervisor \
+                                           python-jinja2 \
+                                           nagios-plugins-all \
+                                           sendEmail \
+                                           perl-Net-SSLeay \
+                                           perl-IO-Socket-SSL && \
```
rhcarvalho: Adding the `rpm -V` line as I mentioned in https://github.com/feedhenry/nagios3-container/pull/2/files/e779ee711ceb759ddd3ffb7c9c8ce3ac3ca0404c#r70609570 prevents us from building bad images (accidentally missing packages).

make-nagios-commands-cfg.py
rhcarvalho: Taking for example the [MySQL image](https://github.com/sclorg/mysql-container/tree/master/5.5):

```console
$ docker run --rm -it \
    -e MYSQL_USER=user \
    -e MYSQL_PASSWORD=pass \
    -e MYSQL_DATABASE=db \
    -p 3306:3306 \
    openshift/mysql-55-centos7
...
160714  6:34:24 [Note] /opt/rh/mysql55/root/usr/libexec/mysqld: ready for connections.
Version: '5.5.45'  socket: '/var/lib/mysql/mysql.sock'  port: 3306  MySQL Community Server (GPL)
```

But without the configuration:

```console
$ docker run --rm -it -p 3306:3306 openshift/mysql-55-centos7
You must either specify the following environment variables:
  MYSQL_USER (regex: '^[a-zA-Z0-9_]+$')
  MYSQL_PASSWORD (regex: '^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$')
  MYSQL_DATABASE (regex: '^[a-zA-Z0-9_]+$')
Or the following environment variable:
  MYSQL_ROOT_PASSWORD (regex: '^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$')
Or both.
Optional Settings:
...
```

We may pass in dummy config, but the image itself never make those choices itself.
Doing so, how will users of the image realize they won't get any emails when things go wrong?!

rhcarvalho:
@mikenairn, knowing that you're taking this from another image, just wanted to hear your opinion on naming executable scripts. Totally subjective topic.

I'd advocate for not including extensions such as ".py" or ".sh". The executable is not tied to a certain programming language. Whenever "start.sh" evolves and we want to turn it into a Perl or Python script, we would just change the shebang and bang it works. Now calling a Perl script "start.sh" is odd.
If it was called "/start", then it could be any script, or could even be a compiled C binary or whatever.

In retrospect, I probably stole that idea from https://google.github.io/styleguide/shell.xml?showone=File_Extensions#File_Extensions

WDYT?

make-nagios-commands-cfg.py
mikenairn: I get your point,  I don't like the magic fallback values either, but this is extra work that was not in the scope of what we planned here. I think we should add a new ticket in the backlog and discuss as a team how best to deal with this and look to resolve it in a separate PR.

mikenairn:
@rhcarvalho Dropping the file extensions from the executable scripts seems OK to me. I think we should just do it for the reasons that you've given.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM centos:centos7
+
+EXPOSE 8080
+
+ADD sendEmail-epel-7.repo /etc/yum.repos.d/
+
+RUN yum install -y epel-release && \
+    INSTALL_PKGS="httpd nagios telnet supervisor python-jinja2 nagios-plugins-all sendEmail perl-Net-SSLeay perl-IO-Socket-SSL" && \
+    yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
```
mikenairn: @rhcarvalho I see your point about the fatter layer because of the update. I have changed this now to be more like https://github.com/sclorg/postgresql-container/blob/master/9.5/Dockerfile, including the rpm -V to check the packages are installed.

make-nagios-commands-cfg.py
rhcarvalho: The change would be:

```diff
-smtp_server = os.getenv('SMTP_SERVER', 'localhost')
+smtp_server = os.environ['SMTP_SERVER']
```

Accessing a key that doesn't exist will raise an exception. If we wanted to be fancy:

```python
import sys
# ...

try:
    smtp_server = os.environ['SMTP_SERVER']
    # ...
except KeyError:
    print "Missing required environment variable"
    sys.exit(1)
```

For our usage with the template, we'll always be setting those environment variables anyway...

PR #3, opened 2016-07-14
Remove appending of '-service' to service names.

mikenairn:
*  The value in the services array should be the service name to look for in openshift.
* Update README based on repo name change. No specific (wrong or otherwise) version mentioned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,3 @@
-# RHMAP Nagios3 for Docker
+# RHMAP Nagios for Docker
 
-Nagios3 server in a Docker container for RHMAP.
+Nagios server in a Docker container for RHMAP.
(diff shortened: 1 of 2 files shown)


mikenairn:
Verification:

```
docker build -t docker.io/rhmap/nagios4:0.0.1 .
docker run -i -e NAGIOS_USER='nagiosadmin' -e NAGIOS_PASSWORD='password' -e RHMAP_ROUTER_DNS='rhmap.local.feedhenry.io' -p 8888:8080 -t docker.io/rhmap/nagios3:0.0.1
```

Should start up OK and nagios dashboard accessible at 192.168.33.12:8888

Currently we only have 2 service being checked (fh-messaging and fh-metrics) since they are shared between core and mbaas. The 4 checks should all go green.

pmccarthy:
As mentioned here [1], I have created this nagios docker image within my Core projects and can confirm that nagios is returning all green.

[1] https://github.com/fheng/fh-openshift-templates/pull/375

PR #5, opened 2016-07-14
RHMAP-8410 install oc tools

jessesarn:
This adds the OSE client tools which provide oc which will be especially useful for its exec functionality when creating nagios checks with probe into the running pods.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,8 @@ RUN yum install -y epel-release && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
+    curl --retry 999 --retry-max-time 0 -sSL https://github.com/openshift/origin/releases/download/v1.2.0/openshift-origin-client-tools-v1.2.0-2e62fab-linux-64bit.tar.gz | tar xzv && \
+    mv openshift-origin-*/* /usr/bin/ \
     mkdir -p /opt/rhmap/ && \
     sed -i -e 's/Listen 80/Listen 8080/' \
            -e 's|DocumentRoot "/var/www/html"|DocumentRoot "/usr/share/nagios/html"|' \


jessesarn:
@rhcarvalho, @mikenairn Could one of you please review when you get a chance?  My other PR's will be based off this branch since they rely on oc/jq to be installed.

Dockerfile
```diff
@@ -2,10 +2,13 @@ FROM centos:centos7
 
 EXPOSE 8080
 
+RUN curl --retry 999 --retry-max-time 0 -sSL https://github.com/openshift/origin/releases/download/v1.2.0/openshift-origin-client-tools-v1.2.0-2e62fab-linux-64bit.tar.gz | tar xzv && \
+  mv openshift-origin-*/* /usr/bin/
```
mikenairn: @rhcarvalho Does this seem like a reasonable way to install the client?

mikenairn: @rhcarvalho What's the best practise with multiple RUN instructions in a Dockerfile? Is it best to keep them separate producing more layers, or bundle them all up in one RUN instruction? i.e Should this be added to the one below?

Dockerfile
```diff
@@ -2,10 +2,13 @@ FROM centos:centos7
...
+
 ADD sendEmail-epel-7.repo /etc/yum.repos.d/
 
 RUN yum install -y epel-release && \
-    INSTALL_PKGS="httpd nagios telnet supervisor python-jinja2 nagios-plugins-all sendEmail perl-Net-SSLeay perl-IO-Socket-SSL" && \
+    INSTALL_PKGS="httpd nagios telnet supervisor python-jinja2 nagios-plugins-all sendEmail perl-Net-SSLeay perl-IO-Socket-SSL jq" && \
```
mikenairn: We should keep @vibe13 in the loop when adding new packages here.

Dockerfile
rhcarvalho: > Does this seem like a reasonable way to install the client?

I think its best to install via yum, Andrea will thank us. The package is called openshift-cli, IIRC.

rhcarvalho: > What's the best practise with multiple RUN instructions in a Dockerfile?

The best practice is to combine them as much as possible.
In general, we should look for having the least number of layers possible, so this practice extends to the use of other instructions as well, such as EXPOSE, ENV, ADD, ...

rhcarvalho:
@jessesarn please rebase this and it should be just about adding `oc`, right?
`jq` should be already there now.

jessesarn:
Reduced to just the oc install via curl.

mikenairn:
Looks good to me.

Dockerfile
```diff
@@ -9,6 +9,8 @@ RUN yum install -y epel-release && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
+    curl --retry 999 --retry-max-time 0 -sSL https://github.com/openshift/origin/releases/download/v1.2.0/openshift-origin-client-tools-v1.2.0-2e62fab-linux-64bit.tar.gz | tar xzv && \
+    mv openshift-origin-*/* /usr/bin/ \
```
rhcarvalho: Isn't it missing a `&&` here?

PR #6, opened 2016-07-14
Rhmap 6727 add mongodb checks

jessesarn:


jessesarn:
@vibe13 Can you confirm whether installing python-pip via yum and then installing pymongo via pip will cause any issues for you with productization ?

Dockerfile
```diff
@@ -21,11 +21,15 @@ RUN yum install -y epel-release && \
     chmod -R 777 /supervisord.log /supervisord.pid /var/log/nagios \
                  /etc/httpd /etc/passwd /var/log /etc/nagios /usr/lib64/nagios \
                  /var/spool/nagios /run /usr/share/httpd /usr/share/nagios && \
-    sed -i -e 's|cfg_file=/etc/nagios/objects/localhost.cfg||' /etc/nagios/nagios.cfg
+    sed -i -e 's|cfg_file=/etc/nagios/objects/localhost.cfg||' /etc/nagios/nagios.cfg && \
+    pip install pymongo
```
rhcarvalho: Perhaps we should at least fix the version we're installing?

Dockerfile
```diff
@@ -21,11 +21,15 @@ RUN yum install -y epel-release && \
...
 
 ADD supervisord.conf /etc/supervisord.conf
 ADD make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
-ADD plugins/default/* plugins/core/* /opt/rhmap/nagios/plugins/
+ADD plugins/default/* \
+    plugins/core/* \
+    https://raw.githubusercontent.com/mzupan/nagios-plugin-mongodb/master/check_mongodb.py \
```
rhcarvalho: This scares me :fearful: 

If we want to use the script and the license is suitable, we should at least vendor (copy) it.

fhservices.cfg.j2
```diff
@@ -15,6 +15,11 @@ define command {
        command_line   $USER1$/check_http -H $HOSTADDRESS$ -S
 }
 
+define command {
+  command_name   check_mongodb
+  command_line   /opt/rhmap/nagios/plugins/check_mongodb_wrapper.sh -A $ARG1$ -S $ARG2$ -w $ARG3$ -c $ARG4$
```
rhcarvalho: Why are the arguments passed as `$ARG1$`, `$ARG2$`, etc? Is it jinja2 syntax? Would it be possible to use named arguments instead of positional?

plugins/default/check_mongodb_wrapper.sh
```diff
@@ -0,0 +1,110 @@
...
+NAGIOS_CRIT=2
+NAGIOS_UNKNOWN=3
+
+################################
+# FUNCTIONS
+################################
+usage() {
+ echo "Usage:"
+ echo " -w <warn>    Warning threshold (default 0)"
+ echo " -c <crit>    Critical threshold (default 0)"
```
rhcarvalho: What do the arguments to `-w` and `-c` mean? IOW, what's the `0`?

plugins/default/check_mongodb_wrapper.sh
```diff
@@ -0,0 +1,110 @@
...
+ echo " -A <action>  Action as defined"
+ echo " -S <name>     OpenShift selector name"
```
rhcarvalho: Nit: indentation is not matching the surrounding lines.

plugins/default/check_mongodb_wrapper.sh
```diff
@@ -0,0 +1,110 @@
...
+done
+
+# Ensure all expected variables are populated
+if [[ -z $ACTION ]] || [[ -z $SELECTOR ]] || [[ -z $WARN ]] || [[ -z $CRIT ]]; then
+ usage; exit $NAGIOS_UNKNOWN
+fi
+
+# Use a temp file to store list of pods with their mongodb connection info
+TEMP_FILE=`mktemp`
+oc get pod --selector=name=$SELECTOR --output json | jq '.items[] | select(.status.phase == "Running") | [.metadata.name, .status.podIP, .spec.containers[].ports[].containerPort, ( .spec.containers[].env[] | select(.name == "MONGODB_ADMIN_PASSWORD") | .value) ] | @tsv' -r > $TEMP_FILE
```
rhcarvalho: I'd be in favor of splitting long lines, specially when there are important things towards the end of the line that could pass unnoticed.

plugins/default/check_mongodb_wrapper.sh
```diff
@@ -0,0 +1,110 @@
...
+# Iterate through each mongodb member, run the check and collect the results 
+while IFS='' read -r line || [[ -n "$line" ]]; do
+  POD=`echo $line | awk '{print $1}'`
+  IP=`echo $line | awk '{print $2}'`
+  PORT=`echo $line | awk '{print $3}'`
+  MONGO_PASS=`echo $line | awk '{print $4}'`
+
+  RESULT_ARR[$CURR]="$POD - `/opt/rhmap/nagios/plugins/check_mongodb.py -H $IP -P $PORT -C $CRIT -W $WARN -A $ACTION -u admin -p $MONGO_PASS 2>&1`"
+
+  CURRENT_RESULT=$?
```
rhcarvalho: This script doesn't set the common options*, and makes use of `$?` to get the exit code of `check_mongodb.py` (I hope that is, and will continue to be, the last command :) ).

I find it to be error prone, specially alarming for something that will be our nagios check.

For instance, there's a `oc get pod` call above, that may fail and the script will continue running.

plugins/default/check_mongodb_wrapper.sh
```diff
@@ -0,0 +1,110 @@
...
+CURR=1
+
+# Iterate through each mongodb member, run the check and collect the results 
+while IFS='' read -r line || [[ -n "$line" ]]; do
+  POD=`echo $line | awk '{print $1}'`
+  IP=`echo $line | awk '{print $2}'`
+  PORT=`echo $line | awk '{print $3}'`
+  MONGO_PASS=`echo $line | awk '{print $4}'`
+
+  RESULT_ARR[$CURR]="$POD - `/opt/rhmap/nagios/plugins/check_mongodb.py -H $IP -P $PORT -C $CRIT -W $WARN -A $ACTION -u admin -p $MONGO_PASS 2>&1`"
```
rhcarvalho: More Bash alarming notes:

If any of those global variables are empty, we may call the python script with funny (missing) arguments like:

```
check_mongodb.py -H -P ...
```

We should have quotes, or not use Bash :)

plugins/default/check_mongodb_wrapper.sh
```diff
@@ -0,0 +1,110 @@
...
+COUNT_CRIT=0
+COUNT_OK=0
+CURR=1
+
+# Iterate through each mongodb member, run the check and collect the results 
+while IFS='' read -r line || [[ -n "$line" ]]; do
+  POD=`echo $line | awk '{print $1}'`
+  IP=`echo $line | awk '{print $2}'`
+  PORT=`echo $line | awk '{print $3}'`
+  MONGO_PASS=`echo $line | awk '{print $4}'`
```
rhcarvalho: We don't need `awk`, `read` can set the values of multiple variables:

```bash
while read POD IP PORT MONGO_PASS; do
  ...
```

plugins/default/check_mongodb_wrapper.sh
```diff
@@ -0,0 +1,110 @@
...
+  elif [ $CURRENT_RESULT -eq $NAGIOS_WARN ]; then
+    COUNT_WARN=$((COUNT_WARN + 1))
+    STATUS="WARN"
+  else
+    COUNT_OK=$((COUNT_OK + 1))
+    STATUS="OK"
+  fi
+
+  CURR=$((CURR + 1))
+done < $TEMP_FILE
```
rhcarvalho: Why do we need the `TEMP_FILE`?!

We could pipe the output of that `oc get` straight into the `while` loop...

fhservices.cfg.j2
jessesarn: The ARG1, ARG2, etc... are nagios conventions of how it handles the parameters inside it's config.  Those are populated by the parameters separated by !'s in the service's check_command.  I'm turning them into named parameters for the calll to the actual script, we can't make it simpler earlier than here, unfortunately.

plugins/default/check_mongodb_wrapper.sh
jessesarn: The warning and critical parameters are threshold values and are more useful for the other check actions in check_mongodb.py script I'm calling.  For connect and replset_state, they don't use thresholds so they expect 0's for both values.  The wrapper could be called for the other types of actions that the underlying script supports, so I am passing them in just in case we use the other features.

plugins/default/check_mongodb_wrapper.sh
jessesarn: This is a trick to workaround a limitation of bash.  A while loop in bash is run as a subprocess and therefore won't have access to track the global variables for the status counts as they'll be reset for each iteration.

plugins/default/check_mongodb_wrapper.sh
jessesarn: Good point, will address.

plugins/default/check_mongodb_wrapper.sh
jessesarn: This was due to my copying and pasting from some of our existing checks in fhcap and not realizing we have inconsistent indentation there.  Fixed to 2 spaces across the script.

plugins/default/check_mongodb_wrapper.sh
jessesarn: Actually, I realized I explained the problem wrong.  It's not the while loop that causes the subprocess, its the piping from the jq on the oc command that causes the subprocess.  If we weren't using a pipeline, we could just put the oc command in the done (i.e. done < `oc ...`).

plugins/default/check_mongodb_wrapper.sh
jessesarn: I've added a counter for unknown statuses and instead of just writing the OK result as the else of the if block, I'm now confirming that there were the same number of OK's as there are instances being checked.  This ensures that if anything goes wrong which affects a check result, we end up with an overall status of NAGIOS_UNKNOWN.

plugins/default/check_mongodb_wrapper.sh
jessesarn: Will do

Dockerfile
jessesarn: So the BSD license which that script is released under can pretty much be included in any other license so long as the copywrite is maintained.  I'll include the file as is in our current code, but if we end up going the python route for our nagios checks, it would probably be much easier to just implement the checks we want directly instead of having a python script calling another python script via shell command.

plugins/default/check_mongodb_wrapper.sh
rhcarvalho: Look:

```console
$ cat while.sh 
#!/bin/bash

G=0

while read a b c; do
  echo $a
  echo $b
  echo $c
  G=$(($G+1))
done < <(yes I love Bash | head -3)

echo $G
$ ./while.sh 
I
love
Bash
I
love
Bash
I
love
Bash
3
```


The problem would be if `while` was part of the pipeline, such as in:

```bash
yes I love Bash | head -3 | while ...
```

In the examples above, replace `yes` with `oc` and `head` with `jq` and we get back to our nagios check ;)

rhcarvalho: But man, the more "tricks", the more we approach write-only code.
Let's not try to do what I propose, but wait and look at a Python alternative, so we can compare.

plugins/default/check_mongodb_wrapper.sh
rhcarvalho: Sorry, I see that in my previous comment I forgot to complete the reference to *.

*
```bash
set -o errexit
set -o nounset
set -o pipefail
```

My concern here is that without those, the script continues running and calling out commands, even though previous commands failed, and we might be working with invalid data.



> This ensures that if anything goes wrong which affects a check result, we end up with an overall status of NAGIOS_UNKNOWN.

I haven't seen this in code. Past line 53 where we validate the input, there is no reference to NAGIOS_UNKNOWN.

plugins/default/check_mongodb_wrapper.sh
rhcarvalho: Does it make sense to call

```
check_mongodb_wrapper.sh - w 42
```

?

If it doesn't, and the argument should always be 0, then we should remove the argument?

rhcarvalho:
@jessesarn now we could progress this by getting rid of `plugins/default/check_mongodb_wrapper.sh` and using the Python code directly.

Could we add a link to where the code was originally taken from?

The we can make changes directly there, perhaps also deleting chunks that we don't use/care.

jessesarn:
@rhcarvalho Please review when you get a chance.  This incorporates the uniform changes you made in (https://github.com/feedhenry/nagios-container/pull/23).

rhcarvalho:
@jessesarn looks like this branch has conflicts with master, could you please rebase?
We can probably get rid of the old commits related to Bash code.

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
...
+from collections import Counter
+
+import openshift
+import nagios
+from mongodb import ReplsetStatuses
+
+
+def generate_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-s', '--selector', type=str, action="store",
```
rhcarvalho: `type=str` and `action="store"` are the defaults, no need to write that :)

I would suggest the "template" from #23:

```python
def generate_parser():
    parser = argparse.ArgumentParser(
        description="Checks the status of a MongoDB replica set",
    )
    parser.add_argument(
        "-s", "--selector", required=True,
        help='label selector for MongoDB pods (e.g.: "name=mongodb")',
    )
    return parser
```

The `required=True` argument is perhaps not "required", it could be okay to filter by no selector (e.g. passing an empty selector == not passing any). The nagios command always pass the argument anyway.


I'm also in doubt about taking a "selector" as input, instead of the service name, more on future comments :)

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
...
+                        required=True, help="Label Selector")
+    return parser
+
+
+check_mongodb_cmd = ("bash", "-c",
+                     "mongo 127.0.0.1/admin "
+                     "-u \"admin\" -p \"$MONGODB_ADMIN_PASSWORD\" "
+                     "--eval=\"rs.status().myState\" --quiet")
```
rhcarvalho: - If you start the list of arguments in the next line, you need less indentation
- Using single quotes allows us to remove backslashes escaping the double quotes
- IMHO mixing multiple arguments (comma separated) and the string concatenation trick (no commas) seem to confuse readers (at least me). It makes one think why are there spaces in the end of the strings, and why are there no commas line in the previous line... and some time later, ah! Implicit string concatenation!

I'd suggest a bit longer line, but clear intention:

```python
check_mongodb_cmd = (
    "bash", "-c",
    'mongo 127.0.0.1/admin -u "admin" -p "$MONGODB_ADMIN_PASSWORD" --eval="rs.status().myState" --quiet'
)
```

WDYT?

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def parse_mongo_result(output):
+    try:
+        return int(output.rstrip("\n"))
+    except:
+        return None
```
rhcarvalho: My brain's type system raised a warning here... why is a function returning integers sometimes and NoneType some other times?

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
+import argparse
+import sys
+import traceback
+from collections import Counter
+
+import openshift
+import nagios
+from mongodb import ReplsetStatuses
```
rhcarvalho: Do we really need this to be defined in another file?

Unless we're planning to have more MongoDB checks that need to know about ReplsetStatuses and could not import it from `mongodb_health`, there's no need for `mongodb.py` and `enum.py`, as there are very specific to this check.

Think if this was Bash... we probably wouldn't start breaking files into a lot of small files and sourcing them from other scripts.

I'd argue that when we look at the list of files under `plugins/default/lib`, we want to see a list of checks, and any noise there is just distracting. I'd strive to keep the noise at a minimum.
Apart from checks, we have there:

- Tests (those our brains can filter out for they all match `test_*.py`)
- `nagios.py` (should be used/imported by all checks)
- `openshift.py` (should be used by most checks)

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def analize(status):
+    if status is None:
+        return nagios.UNKNOWN
+    elif status == ReplsetStatuses.PRIMARY or status == ReplsetStatuses.SECONDARY:
+        return nagios.OK
+    elif status == ReplsetStatuses.ARBITER or status == ReplsetStatuses.STARTUP:
+        return nagios.WARN
+    else:
```
rhcarvalho: - No need for `else` nor the `el` of `elif`, for we're returning in the if blocks.
- It is strange that we say "for anything else return CRITICAL" -- I'd say the logical choice here is "UNKNOWN"? IOW, check explicitly for what makes it OK, WARNING and CRITICAL, everything else is UNKNOWN.


```python
if status in (PRIMARY, SECONDARY):
    return nagios.OK
if status in (ARBITER, STARTUP):
    return nagios.WARN
if ...:
    return nagios.CRIT
return nagios.UNKNOWN
```

Or using a dict we get rid of all ifs:

```python
def analize(status):
    return {
        PRIMARY: nagios.OK,
        SECONDARY: nagios.OK,
        ARBITER: nagios.WARN,
        STARTUP: nagios.WARN,
        ...: nagios.CRIT,
    }.get(status, nagios.UNKNOWN)
```

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
...
+        return nagios.CRIT
+
+
+def report(pods, rs_statuses, nag_statuses):
+    if len(filter(None, pods)) < 1:
```
rhcarvalho: Why do we need to filter pods?

I'd think it is `openshift.get_running_pod_names`'s responsibility to don't return things like `["", "", "foo", ""]`.

rhcarvalho: And when you say `< 1` it really means `== 0`, since length is never negative :)

And since we're saying `== 0`, the pythonic way is to simply say:

```python
if not pods:
    ...
```

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
...
+        return nagios.UNKNOWN
+
+    unique_rs_statuses = Counter(rs_status for rs_status in rs_statuses)
+    unique_nag_statuses = Counter(nag_status for nag_status in nag_statuses)
+
+    ret = max(unique_nag_statuses)
+
+    if unique_rs_statuses[ReplsetStatuses.PRIMARY] != 1:
+        ret = nagios.CRIT
+        print "%s: There are %s nodes claiming to have the primary replica set" % (
```
rhcarvalho: s/to have the primary replica set/to be primary members of the replica set

rhcarvalho: BTW, should not happen with our template, but we're not even checking that the pods that we select are part of the same replica set :)

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,94 @@
...
+            pod,
+            ReplsetStatuses.reverse_mapping[rs_status])
+
+    return ret
+
+
+def check(selector):
+    project = openshift.get_project()
+
+    pods = openshift.get_running_pod_names(project, "name="+selector)
```
rhcarvalho: `"name="` should not be hard coded here, it is part of the selector. A selector has the form `"foo=bar"`.

I suggest using a keyword argument for clarity:

```python
pods = openshift.get_running_pod_names(project, selector=selector)
```

As we talked on IRC, if there would be other flags that we want to pass to `get_running_pod_names`, we'd add in more keyword arguments, instead of relying on positional argument order.

plugins/default/lib/test_mongodb_health.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+import unittest
+
+from mongodb_health import analize, parse_mongo_result, report
+
+
+class TestAnalize(unittest.TestCase):
+
+    def runTest(self):
+        self.assertEqual(analize(1), 0)
```
rhcarvalho: Please don't use magic numbers, use the defined constants.

plugins/default/lib/test_mongodb_health.py
```diff
@@ -0,0 +1,67 @@
...
+            (0, 0)), 1)
+        self.assertEqual(report(
+            ("mongodb-1-1-9hj1x", "mongodb-2-1-xu70c", "mongodb-3-1-wdq55"),
+            (0, 2, 2),
+            (1, 0, 0)), 2)
+        self.assertEqual(report(
+            ("mongodb-1-1-9hj1x", "mongodb-2-1-xu70c", "mongodb-3-1-wdq55"),
+            (1, 7, 2),
+            (0, 1, 0)), 1)
+        self.assertEqual(report((), (), ()), 3)
```
rhcarvalho: Again, please use constants and no magic numbers... I suspect `3` is really `nagios.UNKNOWN`.
Imagine you or someone else looking at a test failure of this test... and you see the error message, you see something like `2 != 3`, and you have no clue what's going on.

We should strive to make the tests legible for new people coming into the code base, not require them to imagine that Nagios use 3 to mean unknown exit code.

plugins/default/lib/test_mongodb_health.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertEqual(parse_mongo_result(""), None)
+
+
+class TestReport(unittest.TestCase):
+
+    def runTest(self):
+        self.assertEqual(report(
+            ("mongodb-1-1-s1ngl",),
+            (1,),
+            (0,)), 0)
```
rhcarvalho: The choice of indentation of these assertions makes it harder to understand what is the input to a function call, and what is the expected value.

Suggestion:

```python
        self.assertEqual(
            report(
                ("mongodb-1-1-s1ngl",),
                (1,),
                (0,),
            ), nagios.OK)
```

What does that `(1,)` and `(0,)` mean?

Maybe I'd even suggest using keyword arguments here to make it a bit clearer:

```python
        self.assertEqual(
            report(
                pods=("mongodb-1-1-s1ngl",),
                rs_statuses=(PRIMARY,),
                nagios_statuses=(nagios.OK,),
            ), nagios.OK)
```

And now, after the rewrite, it makes me think whether this report should really result in `nagios.OK`.
It means that we have a replica set with only one primary and no secondaries... for me the output here should be at least `nagios.WARN`.

plugins/default/lib/mongodb_health.py
jessesarn: The reason for using the selector vs. using the service name is because we use service names in the core vs. mbaas, but the selector is the same in both cases.

@david-martin, @maleck13 - Is the mongodb service name different between an mbaas and core intentional?

plugins/default/lib/mongodb_health.py
jessesarn: Yeah, I had that at one point, but I wasn't sure about mixing double quoting and single quoting for different parameters on the same call, but I do like not having to do the escaping.  As for the length, I only broke it up because of flake8's complaint.  I'll go with this and just ignore any complaint from flake8.

plugins/default/lib/mongodb_health.py
david-martin: If the service name is different, maybe it should be made the same for both. I don't know if it was intentional. @grdryn  Can you better answer @jessesarn question? Perhaps the goal is to realign them as part of (private issue) https://issues.jboss.org/browse/RHMAP-7960

plugins/default/lib/mongodb_health.py
jessesarn: I need a way to differentiate when the command doesn't return a replica set status code.  So anything non numeric is being converted to a None.  If we weren't breaking up the parse, analyze, and report, it would be easier to handle the special cases directly.  I could return something else like -1, but I don't like the idea of creating a new number with special meaning for somebody else's implementation.

Our current methodology for running exec in multiple pods in a single call to openshift.exec_in_pods hides our ability to track the individual return codes from each execution.

plugins/default/lib/mongodb_health.py
jessesarn: If we were monitoring other mongodb replica sets outside of our own, I would agree that we would probably need to do some additional validation; however, that complexity doesn't add us any value in this scenario.  If there are multiple disjointed replica sets, our check would still catch these with its more than 1 primary member check.  Unless of course there was a primary from one, and only secondaries from others getting checked, but the likeliness so is so small and shouldn't be happening at all in our own controlled RHMAP projects.

plugins/default/lib/test_mongodb_health.py
jessesarn: We had discussed this before when @mikenairn brought this up and I thought the decision was to not take advantage of the constants in the tests, but I would agree they would improve readability greatly so happy to do it.

plugins/default/lib/mongodb_health.py
rhcarvalho: If service names are different or not, it doesn't change the implementation of this check :)
It would still take an argument.

rhcarvalho:
@jessesarn we also need to discuss whether it is necessary to connect to each pod, instead of connecting to the replica set and querying the state from there.

I think tests like "are there two primaries" are really testing MongoDB's implementation.

What I've seen useful is to connect to any one pod, and get the replica set status, and make sure we have PRIMARY, SECONDARY, SECONDARY (for the three member template -- so number of members would be an argument), and no errors.

plugins/default/lib/mongodb_health.py
rhcarvalho: The flake8 config is set to be happy with < 120 columns. What I suggested should work (105 cols) without complaints.

plugins/default/lib/mongodb_health.py
rhcarvalho: What is `parse_mongo_result` intended to consume as argument, and what is it intended to return?

From the name, I guess the argument called "output" is a "mongo result", and probably a string, because we call a string method on it.

Now it's not clear what it returns, maybe a number, maybe None, though I don't know what it is semantically.

> Our current methodology for running exec in multiple pods in a single call to openshift.exec_in_pods hides our ability to track the individual return codes from each execution.

Perhaps we need to improve `openshift.exec_in_pods` to allow us do what we need?

plugins/default/lib/mongodb_health.py
rhcarvalho: If we are talking about testing things that never happen, we shouldn't have this check for multiple primaries, for it is not how MongoDB works. We are not testing that MongoDB have a broken implementation of replica sets :)

plugins/default/lib/test_mongodb_health.py
rhcarvalho: [https://github.com/feedhenry/nagios-container/blob/master/plugins/default/lib/test_component_health.py#L13](https://github.com/feedhenry/nagios-container/blob/1bc02dbd22db2741bb726944d95b43f9e796258b/plugins/default/lib/test_component_health.py#L13)

plugins/default/lib/test_mongodb_health.py
jessesarn: I like the naming parameters, so will do this as well.  As far as the result from a single node mongo, unfortunately when it comes to automatically created openshift mbaases, it only includes a single node and so does our core template currently.  I do feel like we could justify warning on this, but I found myself thinking that if we do that we should also warn when a mongo pod doesn't have persistent storage, but that opens a whole new can of worms beyond the point of this check.  Checked with the rest of team and seems to be agreement that single node should return nagios.OK at least for now.

plugins/default/lib/mongodb_health.py
jessesarn: I've adjusted the check to use a service name as input instead which is then used to lookup the selector(s) from the service definition and then those are passed into openshift.get_running_pod_names.  I also modified the function to support multi-pair selectors since the openshift service allows for it.

```
$ oc get service mongodb-service-1b -o json
{
    "kind": "Service",
    "apiVersion": "v1",
... 
   "spec": {
....
        "selector": {
            "foo": "bar",
            "name": "mongodb-replica"
        },
...
```

plugins/default/lib/mongodb_health.py
jessesarn: The output is output from the exec command, its not guaranteed that we have a mongodb result until the parsing has processed the output.  I can't use the output directly as-is because the output includes trailing new line and it shouldn't be the exec that removes that as the new-line could very well mean something in a later check.

I do think we could improve on what openshift.exec_in_pods provides us, and perhaps we should create a JIRA for this that we can bring in at the end of a sprint when we have some extra capacity.  It shouldn't hold up completely this check.  The more checks we create, the more we'll be able to decide what enhancements we can make the shared openshift module functions.

plugins/default/lib/mongodb_health.py
jessesarn: I've merged all of the sub modules into check_mongodb.py.  We can break it out again if we end up writing more checks that could benefit from them.  I do think the nagios module could benefit from being re-written as an enum like I did for the mongo replset statuses, but that can be something else we improve later.

plugins/default/lib/mongodb_health.py
jessesarn: It can happen in a split brain scenario especially with an even number of nodes.  There could be two pairs of PRIMARYs and associated SECONDARYs.

rhcarvalho: This is what MongoDB does in case of a network partition:
https://docs.mongodb.com/manual/core/replica-set-elections/#network-partition

jessesarn:
@rhcarvalho Based on the other conversations going on here.  I'm going to switch to using a service name based approach rather than using a selector, and add an additional check by connecting to the service to query the status of all replset members.  This will complement the existing data already being checked.  I don't have that check added yet, but verifying the other pieces first and going to squash commits and rebase, and then I'll add that tomorrow.

I should be able to exec a command to one of the contributing pods which will connect to the service name instead and get the statuses of all of the mongo members, if the numbers match, then we know we've got a valid consistent replica set for the given service:
```
cfg = rs.status(); res = new Array(); for (var i = 0; i < cfg.members.length; i++) { res[i] = cfg.members[i].state }; res
```
sample output of the above for a 3 node set:
```
[ 1, 2, 2 ]
```

jessesarn:
After deploying a new core based on current version of the scripts, I've discovered a problem with using the service name to find the mongodb pods as well because now we have each mongodb pod creating its own service with its own unique selector.  So I'll need to modify this to discover them a different way which I think is going to be based on the docker image name instead.

fhservices.cfg.j2
```diff
@@ -121,10 +126,18 @@ define service {
...
        use generic-service
-
+       hostgroup_name core,mbaas
        notes One or more pods have volumes which are running out of disk.  You should clean up the files on this pod, if possible, or scale down/up the pod so that it has a fresh filesystem.
        contact_groups rhmapadmins
 }
+
+define service {
+       service_description Database::MongoDB::Health
+       check_command check_mongodb!mongodb-service-1
```
philbrookes: @jessesarn This is failing on my vagrant machine as the mongo service is called `mongodb-1`, I redeployed with the latest fh-openshift template and it is still called `mongodb-1`. Does this need to be renamed?

jessesarn: See my comment: [https://github.com/feedhenry/nagios-container/pull/6#issuecomment-237584827].  With only one pod, its named mongodb-1, but with a 3 node mongodb replica set, we have 3 services now, mongodb-1, 2, and 3, each with a selector for a single pod.

jessesarn:
OK, we're ready for one last review.  This logic is somewhat more complicated, but it covers all of our scenarios known to date for service/pod configurations/layouts.

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,129 @@
...
+        REPLSET_STATUS.DOWN: nagios.CRIT,
+        REPLSET_STATUS.ROLLBACK: nagios.CRIT,
+        REPLSET_STATUS.REMOVED: nagios.CRIT
+    }.get(status, nagios.UNKNOWN)
+
+
+def report(pods, rs_statuses, nag_statuses):
+    if not pods:
+        print "%s: Unable to locate any pods running mongodb" % (
+            nagios.status_code_to_label(nagios.UNKNOWN),)
```
philbrookes: is this a misplaced comma?

philbrookes:
:+1: verified locally

plugins/default/lib/mongodb_health.py
jessesarn: Yeah this is actually intentional.  @rhcarvalho pointed out that when using string prints with variables using this syntax, you should use a tuple even for a single value, this comma ensures that it creates a tuple (which might not happen if the result was a numeric).

rhcarvalho: This is one funny bit of Python syntax, `()` is an empty tuple, `(x, y)` is a 2-element tuple, but `(p)` is not a tuple -- it is just a parenthesized expression. To write a literal 1-element tuple the comma is required: `(p,)`.

plugins/default/lib/mongodb_health.py
```diff
@@ -0,0 +1,129 @@
...
+    args = generate_parser().parse_args()
+    code = nagios.UNKNOWN
+    try:
+        code = check(args.containers)
+    except:
+        traceback.print_exc()
+    finally:
+        sys.exit(code)
+
+# TODO: add logic to validate the health of all members in the json returned for each mongo
```
rhcarvalho: @jessesarn if it makes sense, please create a ticket to track this.

PR #7, opened 2016-07-14
Rhmap 7955 nagios disk checks

jessesarn:


README.md
```diff
@@ -1,3 +1,39 @@
...
+
+```
+cat > nagios-sa.yaml << EOF
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: nagios
+EOF
+ 
+oc create -f nagios-sa.yaml
```
rhcarvalho: You can replace all of the above with:
```bash
oc create sa nagios
# same as
oc create serviceaccount nagios
```

README.md
```diff
@@ -1,3 +1,39 @@
...
+sudo oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:<rhmap-project>:nagios
```
rhcarvalho: No `sudo`.

Instead, you should be logged in as a user with cluster-admin role.

README.md
```diff
@@ -1,3 +1,39 @@
...
+oc policy add-role-to-user admin system:serviceaccount:<rhmap-project>:nagios
+```
+
+2. Edit the nagios deployment configuration to use the nagios service account by adding the serviceAccount and serviceAccountName under spec.template.spec:
```
rhcarvalho: Why is this step required?

rhcarvalho:
@jessesarn could we document why do we need each of the roles that we require for the nagios service account?

- cluster-reader role
- admin role in the project where MBaas/Core is running

I may guess why each of those are needed, but would be nice to have it explained instead of simply asking people to give those permissions.

rhcarvalho:
I have similar concerns with `plugins/default/check_pod_disk_usage.sh` as I had with #6.

Let's have the discussion there.

README.md
jessesarn: Fixed

README.md
jessesarn: This would only be needed if you have an existing deployment config before the nagios service account was introduced.  Our template for the infra nodes should be including this when it creates the app.

rhcarvalho: There's no release yet, so there should be no cluster in that situation.
If that was the case, then this step should be prefixed with the same disclaimer you gave, "This is only needed if ...".

Since it's a very vague step based on `oc edit`, and not necessary for folks installing core today, I'd just get rid of it.

plugins/default/check_pod_disk_usage.sh
```diff
@@ -0,0 +1,106 @@
...
+if [[ -z $THRESH_WARN ]] || [[ -z $THRESH_CRIT ]]; then
+ usage; exit $NAGIOS_UNKNOWN
+fi
+
+TEMP_FILE=`mktemp`
+
+if [ "$CHECK_MODE" = "usage" ]; then
+	DF_CMD="df -x tmpfs"
+elif [ "$CHECK_MODE" = "inodes" ]; then
+	DF_CMD="df -i -x tmpfs"
```
rhcarvalho: Why don't we use the `--output` flag of `df` to get the fields we care about? We could get all fields in one call to `df` and have a single check, instead of calling this script twice with different values of CHECK_MODE.

plugins/default/check_pod_disk_usage.sh
```diff
@@ -0,0 +1,106 @@
...
+else
+	echo "Unknown check mode specified"
+	usage; exit $NAGIOS_UNKNOWN
+fi
+
+oc get pod -o json | jq -r '.items[] | select(.status.phase == "Running") | .metadata | [.namespace, .name] | @tsv' | sort | while IFS=; read podns; do
+	ns=`echo $podns | awk '{print $1}'`
+	pod=`echo $podns | awk '{print $2}'`
+	oc exec $pod -n $ns -- bash -c "$DF_CMD | grep -v Filesystem" | while read line; do
```
rhcarvalho: If for whatever reason there is a mount point called `/foo/bar/Filesystem`, this `grep` call will exclude it from the checks...

plugins/default/check_pod_disk_usage.sh
```diff
@@ -0,0 +1,106 @@
...
+done
+ 
+if [[ -z $THRESH_WARN ]] || [[ -z $THRESH_CRIT ]]; then
+ usage; exit $NAGIOS_UNKNOWN
+fi
+
+TEMP_FILE=`mktemp`
+
+if [ "$CHECK_MODE" = "usage" ]; then
+	DF_CMD="df -x tmpfs"
```
rhcarvalho: Why do we exclude `tmpfs`?
This was making sense to me, until I thought "What if my container is running out of disk space in a tmpfs disk, shouldn't we get a warning for that?"

plugins/default/check_pod_disk_usage.sh
rhcarvalho: Looks like this is what we care about:

```console
[root@7685e0a9afc6 /]# df --output=target,pcent,ipcent | tail -n+2
/                1%    1%
/dev             0%    1%
/sys/fs/cgroup   0%    1%
/run/secrets     0%    1%
/etc/hosts      19%    8%
/dev/shm         0%    1%
```

rhcarvalho: Or, if we want to capture only the lines that match our expected output:

```
df --output=target,pcent,ipcent | grep -P '([\w/]+)\s+(\d+)%\s+(\d+)%'
```

plugins/default/check_pod_disk_usage.sh
```diff
@@ -0,0 +1,106 @@
...
+if [ $COUNT_CRIT -gt 0 ]; then
+	echo "CRIT: Disk $CHECK_MODE on $COUNT_CRIT volumes are over the critical threshold"
+	printf '%s\n' "${RESULT_ARR[@]}"
+	exit $NAGIOS_CRIT
+elif [ $COUNT_WARN -gt 0 ]; then
+	echo "WARN: Disk $CHECK_MODE on $COUNT_WARN volumes are over the warning threshold"
+	printf '%s\n' "${RESULT_ARR[@]}"
+	exit $NAGIOS_WARN
+else
+	echo "OK: Disk $CHECK_MODE on all $COUNT_OK volumes are OK"
```
rhcarvalho: Are we going to return OK if `df` lists nothing?

fhservices.cfg.j2
```diff
@@ -74,3 +79,15 @@ define service {
        contact_groups rhmapadmins
 }
 {% endfor %}
+
+{% for chk in ['usage','inodes'] %}
+define service {
+       service_description Storage::Pod::Disk {{ chk | capitalize() }}
+       hosts {{ rhmap_router_dns }}
+       check_command check_pod_disk_usage!{{ chk }}!80!90
```
rhcarvalho: Is there an use case to pass in different percentages for `inodes` and `usage` modes?
Asking because if we check both block usage and inode usage in one single check, then possibly the threshold argument would be shared.

I saw in a container output like this:

```
Mounted on     Use% IUse%
/etc/hosts      19%    8% 
```

So those percentages are not necessarily in sync.

jessesarn: Yeah.  I'm with you.  Our current checks in our SaaS platform are separate ones, but I see no reason why we can't get the data for both pieces in one check.  We should probably keep separate thresholds for parameters just in case we decide to do it differently.

plugins/default/check_pod_disk_usage.sh
jessesarn: I tried with --output and had some weird results, but I honestly can't remember what they were, but may have been when I was having problems getting oc exec to let me pass parameters which I later figured out how to do properly.

jessesarn: Oh, I figured out why, I wasn't using an =.  I like the cleanliness here.

plugins/default/check_pod_disk_usage.sh
jessesarn: We control what our mount points are on our components, so I wouldn't think we would ever allow that to be the case, but I see no harm in using tail to avoid the header line like you did in your python based PR.

plugins/default/check_pod_disk_usage.sh
jessesarn: Based on the filesystems I've seen in our existing pods, all of the tmpfs ones wouldn't be getting data written to them normally, but I guess its possible something could be doing evil stuff it shouldn't.  We can include them, but the output listing returned to nagios is going to be awfully large.  Maybe its best to include them for now and maybe we could trim things down later once we get an idea of what it looks like.

plugins/default/check_pod_disk_usage.sh
jessesarn: I've change the mongo check to only return OK if the number of OK's matches the number of results being parsed.  We should do the same here which I think would make this a non issue.

jessesarn:
@rhcarvalho I know its rough, but at least it finally get this working.  I couldn't figure out the right way to pass multiple parameters into map, so I resorted to two for loops that you might be able to help clean up.  The problems I see are:

1. the number of tests when running the unittests doesn't seem right. It says it ran 6 tests, but I currently have 11, but they do seem to run.
2. our ADD statement in the Dockerfile now copies all of the files in the lib subfolder during the copy of the plugins/core, plugins/default folders into /opt/rhmap/nagios/plugins.  So we are currently getting dupes of all those files.  Also the symlinks for the checks are ending up pointing to the target file (i.e. disk-usage is supposed to link to lib/disk-usage.py, but instead its a full copy of the file)
3. Running the script in the container goes horribly bad even though I can run the script with my local python and oc available.   When run in the container:

```
./disk_usage.py -w 80 -c 90
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x587f85]

goroutine 1 [running]:
panic(0x1bab300, 0xc820014080)
	/usr/local/go/src/runtime/panic.go:464 +0x3e6
github.com/openshift/origin/pkg/cmd/cli/cmd.ProjectOptions.RunProject(0x0, 0x0, 0x0, 0x0, 0x0, 0xc820040ea0, 0xc820040ed0, 0xc820040f00, 0xc820040f30, 0x0, ...)
	/go/src/github.com/openshift/origin/pkg/cmd/cli/cmd/project.go:129 +0xc45
github.com/openshift/origin/pkg/cmd/cli/cmd.NewCmdProject.func1(0xc820116600, 0xc8203fecf0, 0x0, 0x1)
	/go/src/github.com/openshift/origin/pkg/cmd/cli/cmd/project.go:77 +0x1a3
github.com/spf13/cobra.(*Command).execute(0xc820116600, 0xc8203fec90, 0x1, 0x1, 0x0, 0x0)
	/go/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/spf13/cobra/command.go:572 +0x85a
github.com/spf13/cobra.(*Command).ExecuteC(0xc82029c400, 0xc820116600, 0x0, 0x0)
	/go/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/spf13/cobra/command.go:662 +0x53f
github.com/spf13/cobra.(*Command).Execute(0xc82029c400, 0x0, 0x0)
	/go/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/spf13/cobra/command.go:618 +0x2d
main.main()
	/go/src/github.com/openshift/origin/cmd/oc/oc.go:27 +0x180

```

vs. locally run
```
$ ./disk_usage.py -w 10 -c 15
WARN: There are 9 volumes over the warning threshold
OK: core-monitoring-39-ax8y0:/ - bytes used: 1%, inodes used: 1%
OK: core-monitoring-39-ax8y0:/dev - bytes used: 0%, inodes used: 1%
OK: core-monitoring-39-ax8y0:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: core-monitoring-39-ax8y0:/run/secrets - bytes used: 1%, inodes used: 1%
OK: core-monitoring-39-ax8y0:/etc/hosts - bytes used: 9%, inodes used: 1%
OK: core-monitoring-39-ax8y0:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: core-monitoring-39-ax8y0:/dev/shm - bytes used: 0%, inodes used: 1%
OK: fh-aaa-1-d3j0a:/ - bytes used: 1%, inodes used: 1%
OK: fh-aaa-1-d3j0a:/dev - bytes used: 0%, inodes used: 1%
OK: fh-aaa-1-d3j0a:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: fh-aaa-1-d3j0a:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: fh-aaa-1-d3j0a:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: fh-aaa-1-d3j0a:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: fh-aaa-1-d3j0a:/dev/shm - bytes used: 0%, inodes used: 1%
OK: fh-appstore-1-uulx3:/ - bytes used: 1%, inodes used: 1%
OK: fh-appstore-1-uulx3:/dev - bytes used: 0%, inodes used: 1%
OK: fh-appstore-1-uulx3:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: fh-appstore-1-uulx3:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: fh-appstore-1-uulx3:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: fh-appstore-1-uulx3:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: fh-appstore-1-uulx3:/dev/shm - bytes used: 0%, inodes used: 1%
OK: fh-ngui-1-78rt5:/ - bytes used: 1%, inodes used: 1%
OK: fh-ngui-1-78rt5:/dev - bytes used: 0%, inodes used: 1%
OK: fh-ngui-1-78rt5:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: fh-ngui-1-78rt5:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: fh-ngui-1-78rt5:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: fh-ngui-1-78rt5:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: fh-ngui-1-78rt5:/dev/shm - bytes used: 0%, inodes used: 1%
OK: fh-scm-1-ysso6:/ - bytes used: 1%, inodes used: 1%
OK: fh-scm-1-ysso6:/dev - bytes used: 0%, inodes used: 1%
OK: fh-scm-1-ysso6:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: fh-scm-1-ysso6:/tmp - bytes used: 9%, inodes used: 1%
OK: fh-scm-1-ysso6:/run/secrets - bytes used: 1%, inodes used: 1%
OK: fh-scm-1-ysso6:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: fh-scm-1-ysso6:/dev/shm - bytes used: 0%, inodes used: 1%
OK: fh-supercore-1-fnss6:/ - bytes used: 1%, inodes used: 1%
OK: fh-supercore-1-fnss6:/dev - bytes used: 0%, inodes used: 1%
OK: fh-supercore-1-fnss6:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: fh-supercore-1-fnss6:/run/secrets - bytes used: 1%, inodes used: 1%
OK: fh-supercore-1-fnss6:/etc/hosts - bytes used: 9%, inodes used: 1%
OK: fh-supercore-1-fnss6:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: fh-supercore-1-fnss6:/dev/shm - bytes used: 0%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/ - bytes used: 1%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/dev - bytes used: 0%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: gitlab-shell-1-6c2kd:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/etc/feedhenry/gitlab-shell - bytes used: 1%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/home/git/gitlab-config - bytes used: 1%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: gitlab-shell-1-6c2kd:/dev/shm - bytes used: 0%, inodes used: 1%
OK: httpd-1-oxm0h:/ - bytes used: 1%, inodes used: 1%
OK: httpd-1-oxm0h:/dev - bytes used: 0%, inodes used: 1%
OK: httpd-1-oxm0h:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: httpd-1-oxm0h:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: httpd-1-oxm0h:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: httpd-1-oxm0h:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: httpd-1-oxm0h:/dev/shm - bytes used: 0%, inodes used: 1%
OK: memcached-1-tkszc:/ - bytes used: 1%, inodes used: 1%
OK: memcached-1-tkszc:/dev - bytes used: 0%, inodes used: 1%
OK: memcached-1-tkszc:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: memcached-1-tkszc:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: memcached-1-tkszc:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: memcached-1-tkszc:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: memcached-1-tkszc:/dev/shm - bytes used: 0%, inodes used: 1%
OK: millicore-1-gcwja:/ - bytes used: 3%, inodes used: 1%
OK: millicore-1-gcwja:/dev - bytes used: 0%, inodes used: 1%
OK: millicore-1-gcwja:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: millicore-1-gcwja:/run/secrets - bytes used: 1%, inodes used: 1%
OK: millicore-1-gcwja:/etc/hosts - bytes used: 9%, inodes used: 1%
OK: millicore-1-gcwja:/etc/feedhenry/core - bytes used: 1%, inodes used: 1%
OK: millicore-1-gcwja:/etc/feedhenry/gitlab-shell - bytes used: 1%, inodes used: 1%
OK: millicore-1-gcwja:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: millicore-1-gcwja:/dev/shm - bytes used: 0%, inodes used: 1%
OK: mongodb-1-1-vir3e:/ - bytes used: 1%, inodes used: 1%
OK: mongodb-1-1-vir3e:/dev - bytes used: 0%, inodes used: 1%
OK: mongodb-1-1-vir3e:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: mongodb-1-1-vir3e:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: mongodb-1-1-vir3e:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: mongodb-1-1-vir3e:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: mongodb-1-1-vir3e:/dev/shm - bytes used: 0%, inodes used: 1%
OK: mysql-1-id0hf:/ - bytes used: 1%, inodes used: 1%
OK: mysql-1-id0hf:/dev - bytes used: 0%, inodes used: 1%
OK: mysql-1-id0hf:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: mysql-1-id0hf:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: mysql-1-id0hf:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: mysql-1-id0hf:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: mysql-1-id0hf:/dev/shm - bytes used: 0%, inodes used: 1%
OK: redis-1-kn2v6:/ - bytes used: 1%, inodes used: 1%
OK: redis-1-kn2v6:/dev - bytes used: 0%, inodes used: 1%
OK: redis-1-kn2v6:/sys/fs/cgroup - bytes used: 0%, inodes used: 1%
OK: redis-1-kn2v6:/run/secrets - bytes used: 1%, inodes used: 1%
WARN: redis-1-kn2v6:/etc/hosts - bytes used: 10%, inodes used: 1%
OK: redis-1-kn2v6:/run/secrets/kubernetes.io/serviceaccount - bytes used: 1%, inodes used: 1%
OK: redis-1-kn2v6:/dev/shm - bytes used: 0%, inodes used: 1%
```

rhcarvalho:
> the number of tests when running the unittests doesn't seem right. It says it ran 6 tests, but I currently have 11, but they do seem to run.

@jessesarn I think you're counting the number of assertions. That is different than the number of tests.
A well accepted norm / "good practice" is to have one assertion per test, so that whenever you see a test failure, you know exactly what it was.

This is due to the way how assertions work, they raise exceptions that interrupt the program flow.
Whenever the first assertion in a test method fails, none of the others are executed.

I went with multiple assertions balancing on the cost-benefit of having many methods/lines vs. the value they give. We want all tests passing anyway.

Just for illustration of the Go x Python differences, in Go there are no exceptions. In the testing library, you can mark the test as failed and keep making furthers checks. So a test run may report multiple failure reasons, while in Python and other languages that uses exceptions, a test is limited to a single failure reason.

Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN yum install -y epel-release && \
 ADD supervisord.conf /etc/supervisord.conf
 ADD make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
 ADD plugins/default/* plugins/core/* /opt/rhmap/nagios/plugins/
+ADD plugins/default/lib/* plugins/core/lib/* /opt/rhmap/nagios/plugins/lib/
```
rhcarvalho: We don't need this line, because the line above should copy everything recursively.

Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN yum install -y epel-release && \
 ADD supervisord.conf /etc/supervisord.conf
 ADD make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
 ADD plugins/default/* plugins/core/* /opt/rhmap/nagios/plugins/
```
rhcarvalho: I haven't normally paid attention to that, but we should be preferring to use `COPY` instead of `ADD`:
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/add-or-copy

README.md
```diff
@@ -1,3 +1,32 @@
 # RHMAP Nagios for Docker
 
 Nagios server in a Docker container for RHMAP.
+
+# Setting up nagios service account
+
+In order to talk to the OpenShift and perform exec commands on pods, a service account will need to be created for use by nagios and given the proper permissions.
```
rhcarvalho: I'm still missing some quick notes here, in the README, explaining why those specific permissions are required.

README.md
```diff
@@ -1,3 +1,32 @@
...
+
+1. Create the account using oc in the same project as the mbaas/core you are monitoring:
+
+```
```
rhcarvalho: You need to indent this, otherwise you get two independent numbered lists starting from 1, see:

https://github.com/jessesarn/nagios-container/blob/5f915f9b8bf8fa9efd4597f9e55ad50a6db010c1/README.md

rhcarvalho:
> Also the symlinks for the checks are ending up pointing to the target file (i.e. disk-usage is supposed to link to lib/disk-usage.py, but instead its a full copy of the file)

I reproduced this locally, still haven't found a fix.

@mikenairn @maleck13 @pmccarthy @david-martin perhaps you know some workaround here? We may be forced to revisit the file organization and use of symlinks.

rhcarvalho:
> Running the script in the container goes horribly bad even though I can run the script with my local python and oc available. When run in the container:

@jessesarn the error you posted is from Go code, from the oc command. How did you "run in the container"?

plugins/default/lib/nagios.py
```diff
@@ -3,3 +3,12 @@
...
 CRIT=2
 UNKNOWN=3
+
+def status_code_to_label(code):
+	return {
+      OK: "OK",
+      WARN: "WARN",
+      CRIT: "CRIT",
+      UNKNOWN: "UNKNOWN"
+    }.get(code, UNKNOWN)
```
rhcarvalho: The default value should probably be the string `"UNKNOWN"`, not a reference to the constant?

plugins/default/lib/nagios.py
```diff
@@ -3,3 +3,12 @@
 WARN=1
 CRIT=2
 UNKNOWN=3
+
+def status_code_to_label(code):
+	return {
```
rhcarvalho: Could we run `autopep8` to reformat all Python files?

We should probably document it too :)

https://pypi.python.org/pypi/autopep8

jessesarn:
@rhcarvalho The error from go is apparently during the 'oc project -q'.  When I rsh into the pod and run that command manually, I get the same error.
```
sh-4.2$ oc project
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x587f85]

goroutine 1 [running]:
panic(0x1bab300, 0xc820014090)
	/usr/local/go/src/runtime/panic.go:464 +0x3e6
github.com/openshift/origin/pkg/cmd/cli/cmd.ProjectOptions.RunProject(0x0, 0x0, 0x0, 0x0, 0xc820024c00, 0xc8200aa840, 0xc8200aa870, 0xc8200aa8a0, 0xc8200aa8d0, 0x0, ...)
	/go/src/github.com/openshift/origin/pkg/cmd/cli/cmd/project.go:129 +0xc45
github.com/openshift/origin/pkg/cmd/cli/cmd.NewCmdProject.func1(0xc8200edc00, 0x2e54ca0, 0x0, 0x0)
	/go/src/github.com/openshift/origin/pkg/cmd/cli/cmd/project.go:77 +0x1a3
github.com/spf13/cobra.(*Command).execute(0xc8200edc00, 0x2e54ca0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/spf13/cobra/command.go:572 +0x85a
github.com/spf13/cobra.(*Command).ExecuteC(0xc8200ec600, 0xc8200edc00, 0x0, 0x0)
	/go/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/spf13/cobra/command.go:662 +0x53f
github.com/spf13/cobra.(*Command).Execute(0xc8200ec600, 0x0, 0x0)
	/go/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/spf13/cobra/command.go:618 +0x2d
main.main()
	/go/src/github.com/openshift/origin/cmd/oc/oc.go:27 +0x180
```

I can see it knows the project using other commands.  Maybe I need to just work around it by running oc status and parsing that first line to get the project.
```
sh-4.2$ oc status | head -n1
In project core on server https://172.30.0.1:443
```

Any ideas why that command would be giving such a strange code?

plugins/default/lib/openshift.py
```diff
@@ -27,3 +27,13 @@ def _exec_in_pods(oc, project, pods, cmd):
 
 def exec_in_pods(project, pods, cmd):
     return _exec_in_pods(oc, project, pods, cmd)
+
+
+def get_project():
+    # Manually get the namespace via workaround because of a bug in 'oc project', see:
+    # <insert bug url here>
+    # return openshift.oc("project", "-q").rstrip("\n")
```
rhcarvalho: We should definitely open the bug, but not sure we need this here. The other approach is a documented way to call the API from a pod:
http://kubernetes.io/docs/user-guide/accessing-the-cluster/#accessing-the-api-from-a-pod

plugins/mbaas/README.md
```diff
@@ -0,0 +1,4 @@
+
+Contents in here will overlap on top of the files from the default subtree
+
+*Note: Any files with the same sub path and name as a file from default will be overwritten with the one here (This allows for core/mbaas to override the behavior we have in default)*
```
rhcarvalho: We have a future task to allow using the image for mbaas or core. Isn't this README a sign that maybe we should defer to a different method, and not be merging directory trees?

jessesarn: So this is very similar to what we do with our current SaaS offering and chef.  We have "default" files that can be overriden by ones based on os families and such.  Its unlikely we'll use the overriding concept here, its more so that we can have checks for just the core in particular that won't have any concept in the mbaas.  So rather than have the scripts that are purely core specific getting copied into a nagios pod for mbaas, it can avoid them entirely.  When the plugins/core path didn't exist and I removed the * (which was flatting the files/subfolders), it caused the ADD/COPY to fail because the source didn't exist.  Ironcially, I can remove the README's now since the folder will still stay around since I will have a commit that adds them and a commit that removes them.  I'm happy to take these README's back out.

Dockerfile
```diff
@@ -17,16 +17,16 @@ RUN yum install -y epel-release && \
...
                  /etc/httpd /etc/passwd /var/log /etc/nagios /usr/lib64/nagios \
                  /var/spool/nagios /run /usr/share/httpd /usr/share/nagios && \
     sed -i -e 's|cfg_file=/etc/nagios/objects/localhost.cfg||' /etc/nagios/nagios.cfg
 
-ADD supervisord.conf /etc/supervisord.conf
-ADD make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
-ADD plugins/default/* plugins/core/* /opt/rhmap/nagios/plugins/
+COPY supervisord.conf /etc/supervisord.conf
+COPY make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
+COPY plugins/default/ plugins/core/ /opt/rhmap/nagios/plugins/
```
rhcarvalho: Since this is copying all of the things under `plugins/`, we should copy the whole tree instead of enumerating each child:

```Dockerfile
COPY plugins /opt/rhmap/nagios/
```

Wouldn't that be simpler?

plugins/default/check_pod_disk_usage.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
```
rhcarvalho: We don't need this file anymore, right?

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+    crit = parser.crit
+
+    if crit < warn:
+        raise ValueError("Critical threshold lower than Warning threshold")
+
+    project = openshift.get_project()
+
+    results = []
+
+    for p in openshift.exec_in_pods(project, openshift._get_running_pods(
```
rhcarvalho: The names starting with `_` are to be considered private. We should be using `openshift.get_running_pods(project)`.

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+            openshift.oc, project), check_disk_cmd):
+        for d in analize(p[0], parse_df_lines(p[1]), warn, crit):
```
rhcarvalho: Can we give names to things like `p[0]`, `p[1]`, etc?

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+            disk_status = nagios.WARN
+        elif max_pcent < warning_threshold:
+            disk_status = nagios.OK
+
+        results.append([pod, disk[0], disk[1], disk[2], disk_status])
+    return results
+
+
+def report(results):
+    counts = Counter(d[4] for d in results)
```
rhcarvalho: Can we try to name things better here too, what is `d[4]`?

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
     return (mo.group(1), int(mo.group(2)), int(mo.group(3)))
 
-def analize(pcent, ipcent, warning_threshold, critical_threshold):
-    max_pcent = max(pcent, ipcent)
-    return max_pcent >= warning_threshold, max_pcent >= critical_threshold
 
-def report():
-    raise NotImplementedError("report is not implemented yet")
+def parse_df_lines(lines):
+    return (map(parse_df_line, lines.splitlines()[1:]))
```
rhcarvalho: `parse_df` was already such that it ignored a line that didn't match the pattern, why are we slicing off the first line here?

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+        print "%s: All %s volumes are under the warning threshold" % (nagios.status_code_to_label(ret), len(results))
+    elif ret == nagios.UNKNOWN:
+        print "%s: Unable to determine usage on %s volumes" % (nagios.status_code_to_label(ret), counts[nagios.UNKNOWN])
+    elif ret == nagios.WARN:
+        print "%s: There are %s volumes over the warning threshold" % (nagios.status_code_to_label(ret), counts[nagios.WARN])
+    else:
+        print "%s: There are %s volumes over the critical threshold and %s volumes over the warning threshold" % (nagios.status_code_to_label(ret), counts[nagios.CRIT], counts[nagios.WARN])
+
+    for r in results:
+        print "%s: %s:%s - bytes used: %s%%, inodes used: %s%%" % (nagios.status_code_to_label(r[4]), r[0], r[1], r[2], r[3])
```
rhcarvalho: Please name these `r[n]` things.

plugins/default/check_pod_disk_usage.sh
jessesarn: correct, removing

plugins/default/lib/disk_usage.py
jessesarn: Sorry, that was from when I was debugging why the namespace filter wasn't working which I ended up finding was due to the new line at the end.

plugins/default/lib/disk_usage.py
jessesarn: I'm unpacking the tuple into pod and disks now.

plugins/default/lib/disk_usage.py
jessesarn: Unpacked to name, space_usage, inode_usage

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+import openshift
 import nagios
 
 
 check_disk_cmd = ("df", "--output=target,pcent,ipcent")
 # Example output:
 # /etc/hosts      19%    8%
-check_disk_output_pattern = re.compile(r"([\w/]+)\s+(\d+)%\s+(\d+)%")
+check_disk_output_pattern = re.compile(
+    r"([A-Za-z0-9\./\-_]+)\s+(\d+)%\s+(\d+)%")
```
rhcarvalho: Why did we need to change the pattern?

plugins/default/lib/disk_usage.py
jessesarn: It returns an empty tuple not really ignores it.  I thought it seemed cleaner to remove the line rather than have to post process the results of the map call to filter it out, but I can see how its probably better to avoid just blindly ignoring a line, so I've adjusted to use a list comprehension to remove empty tuple(s).

plugins/default/lib/disk_usage.py
jessesarn: Because it missed several file systems in my testing.  Examples are:
/etc/feedhenry/gitlab-shell (because of the -)
/run/secrets/kubernetes.io/serviceaccount (because of the .)

plugins/default/lib/disk_usage.py
jessesarn: unpacked to pod, name, disk_usage, inode_usage, status

plugins/default/lib/disk_usage.py
rhcarvalho: It was intended to be used with the default `filter` behavior:

```python
>>> filter(None, [(), (1, 2), (3, 4)])
[(1, 2), (3, 4)]
```

jessesarn: thats cleaner, will use filter instead

plugins/default/lib/disk_usage.py
rhcarvalho: I see. Haven't put enough thought on that originally.
Regexps can easily bite us, would it make sense to use a negative pattern / something less restrictive?
E.g., we could assume that paths contain anything but spaces or tabs: `[^ \t]`.

Maybe there's a better way to approach this... I wonder what `df` outputs when the mount path has spaces.
Is `df` separating the fields by just a tab char?

jessesarn: I was going to really work out the perfect filter, but the reality is that we are the ones controlling what's going into these pods in our RHMAP projects, so we shouldn't have to worry about things that are too crazy.  However, since our df commands is returning two easy to regex numbers along with the name, we could just order the results with the %'s first and the name at the end.  Then we could just match the numbers for the first two groups and the name is everything else to the end of the line.  What do you think?

jessesarn: Yeah, putting target as the last field works great.  This expression '(\d+)%\s+(\d+)%\s+(.+)' handles spaces, -'s, .'s and pretty much guarantees everything works.  I'm going to add some of these scenarios to the tests before I push my commits.

rhcarvalho: > since our df commands is returning two easy to regex numbers along with the name, we could just order the results with the %'s first and the name at the end. Then we could just match the numbers for the first two groups and the name is everything else to the end of the line. What do you think?

Brilliant!

jessesarn: Also changed my use of name for unpacked tuples to use mount instead to make it more clear.

fhservices.cfg.j2
```diff
@@ -90,3 +95,13 @@ define service {
...
       contact_groups rhmapadmins
 }
+
+define service {
+       service_description Storage::Pod::Disk Usage
+       hosts {{ rhmap_router_dns }}
+       check_command check_pod_disk_usage!80!90
+       use generic-service
+
+       notes One or more pods have volumes which are running out of disk.  You should clean up the files on this pod, if possible, or scale down/up the pod so that it has a fresh filesystem.
```
rhcarvalho: Is "notes" a command / keyword?

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
 from subprocess import CalledProcessError
+from collections import Counter
 
-from openshift import oc
+import openshift
 import nagios
 
 
-check_disk_cmd = ("df", "--output=target,pcent,ipcent")
+check_disk_cmd = ("df", "--output=pcent,ipcent,target")
```
rhcarvalho: This is a brilliant move! :star2:

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
 # Example output:
 # /etc/hosts      19%    8%
-check_disk_output_pattern = re.compile(r"([\w/]+)\s+(\d+)%\s+(\d+)%")
+check_disk_output_pattern = re.compile(
+    r"\s*(\d+)%\s+(\d+)%\s+(.+)$")
```
rhcarvalho: Any reason to split into 2 lines?

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+def parse_df_line(line):
     mo = check_disk_output_pattern.match(line)
     if mo is None:
         return ()
-    return (mo.group(1), int(mo.group(2)), int(mo.group(3)))
+    return (mo.group(3), int(mo.group(1)), int(mo.group(2)))
+
+
+def parse_df_lines(lines):
+    return (filter(None, map(parse_df_line, lines.splitlines())))
```
rhcarvalho: The outter parens are spurious.

rhcarvalho: Making it look like Lisp code :)

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+
+        results.append([pod, mount, space_usage, inode_usage, disk_status])
+    return results
+
 
-def analize(pcent, ipcent, warning_threshold, critical_threshold):
-    max_pcent = max(pcent, ipcent)
-    return max_pcent >= warning_threshold, max_pcent >= critical_threshold
+def report(results):
+    counts = Counter(d[4] for d in results)
```
rhcarvalho: Please let's not have any magic numbers / indices.
What is `d[4]`? Why is a single result called `d`?!

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+
+    return ret
 
-def report():
-    raise NotImplementedError("report is not implemented yet")
 
 def main():
-    project = oc("project", "-q")
-    print analize(map(parse_df, exec_in_pods(project, get_running_pod_names(), check_disk_cmd)))
+    parser = generate_parser().parse_args()
```
rhcarvalho: This was fixed in a similar PR, the return value is not a parser.

rhcarvalho: s/parser/args

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+
+    if crit < warn:
+        raise ValueError("Critical threshold lower than Warning threshold")
+
+    project = openshift.get_project()
+
+    results = []
+
+    for pod, lines in openshift.exec_in_pods(project, openshift.get_running_pod_names(
+            project), check_disk_cmd):
```
rhcarvalho: This line break here is weird, doesn't help reading the line.

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,97 @@
...
+        for d in analize(pod, parse_df_lines(lines), warn, crit):
```
rhcarvalho: What does `d` mean?

plugins/default/lib/openshift.py
```diff
@@ -5,18 +5,35 @@
 def oc(*args):
     return check_output(("oc",) + args)
 
-def _get_running_pod_names(oc, project):
+
+def _get_running_pods(oc, project):
```
rhcarvalho: Why are we changing the function name?

plugins/default/lib/openshift.py
```diff
@@ -5,18 +5,35 @@
...
     # Manually filter Running pods because of a bug in `oc get`, see:
     # https://github.com/kubernetes/kubernetes/issues/29115
-    #return oc("-n", project, "get", "pods", "--show-all=false", "-o", "name")
-    pods = json.loads(oc("-n", project, "get", "pods", "--show-all=false", "-o", "json"))
-    return [p["metadata"]["name"] for p in pods if p["status"]["phase"] == "Running"]
+    # return oc("-n", project, "get", "pods", "--show-all=false", "-o", "name")
+    pods = json.loads(oc("-n", project, "get", "pods",
+                         "--show-all=false", "-o", "json"))
+    return [p["metadata"]["name"]
+            for p in pods["items"] if p["status"]["phase"] == "Running"]
```
rhcarvalho: Looks like you're fixing a bug :)
I think `oc` is returning a `List` object? We shouldn't call it `pods`.
Fixing that naming problem, this `return ...` line can stay as it was before.

fhservices.cfg.j2
jessesarn: Yes, it is displayed in the nagios service detail and we use it to explain where the most likely cause is (and probably later a URL for our docs with more detail)

plugins/default/lib/disk_usage.py
jessesarn: It wasn't me, it was autopip.  Not really ecstatic about its line length limit as it makes some of these things look really stupid.  Think we have excluded that check in the travis job, so I could put these lines back the way they were and run with the same skip.

plugins/default/lib/disk_usage.py
jessesarn: The last item is the nagios status of each volume check.  All I want to do is get a count based on the unique number of statuses so that I can generate the summary line with the number of each type of problem as well as get the highest nagios code which is the one used for the overall status.  If there is a cleaner way, let me know.

plugins/default/lib/disk_usage.py
jessesarn: The results of the analize return a lists of tuples that have information about each disk.  Since the data from exec comes in as multiple lines per pod, I am denormalizing the data so that each result has the pod name, the disk name, usage numbers, and the status of its threshold check.  This is just taking all those separate items and adding them to a list which has the data across all pods/volumes so that reporting is simply able to dump each row.

plugins/default/lib/openshift.py
jessesarn: I changed it because the original version you had the exported function name was calling the internal function name without the _names in it, I should have fixed the internal name instead of changing the call.  This is now fixed.

plugins/default/lib/openshift.py
jessesarn: it was just autopep breaking up the lines.  I'll put them back together as they were.

fhservices.cfg.j2
rhcarvalho: The advice here is a bit strange, one doesn't normally "clean up the files on this pod".
The pod/container is normally cattle, ephemeral, and disposable; one does no exec into it and try to delete things to free space.

The space problems that uses need to act upon may arise in the nodes and in persistent volumes.

WDYT?

david-martin: Is this stepping on the toes of what this jira is for? https://issues.jboss.org/browse/RHMAP-9026
I don't think it matters too much what's in these notes at this point

plugins/default/lib/disk_usage.py
rhcarvalho: The cleaner way is to give names in the code, rather than comments in the PR :)

When I'm reading, the first thing that makes understanding harder is to imagine what is `d`. A "result" shortened as "r" is probably more reasonable than "d", what does "d" stand for?!

I see `d` is also used for the return value of `analize`.

What is `d[4]`? I have to go find where `results` was defined to figure out what it is... I have to search everywhere where `report` is being called. Now that I found it, I see it is constructed from calling `analize`, so I have to go see what it returns...

And the thing with index 4 is called `disk_status`.

Suggestion:

```python
unique_statuses = Counter(disk_status for pod, mount, space_usage, inode_usage, disk_status in results)
```

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,95 @@
...
+    if crit < warn:
+        raise ValueError("Critical threshold lower than Warning threshold")
+
+    project = openshift.get_project()
+
+    results = []
+
+    for pod, lines in openshift.exec_in_pods(project, openshift.get_running_pod_names(project), check_disk_cmd):
+        for d in analize(pod, parse_df_lines(lines), warn, crit):
+            results.append(d)
```
rhcarvalho: You can remove one `for` loop:

```python
results = []
for pod, lines in ...:
    results.extend(analize(pod, parse_df_lines(lines), warn, crit))
```

plugins/default/lib/openshift.py
```diff
@@ -5,18 +5,31 @@
...
+
 def _get_running_pod_names(oc, project):
     # Manually filter Running pods because of a bug in `oc get`, see:
     # https://github.com/kubernetes/kubernetes/issues/29115
-    #return oc("-n", project, "get", "pods", "--show-all=false", "-o", "name")
-    pods = json.loads(oc("-n", project, "get", "pods", "--show-all=false", "-o", "json"))
-    return [p["metadata"]["name"] for p in pods if p["status"]["phase"] == "Running"]
+    # return oc("-n", project, "get", "pods", "--show-all=false", "-o", "name")
+    pods = json.loads(oc("-n", project, "get", "pods","--show-all=false", "-o", "json"))
+    return [p["metadata"]["name"] for p in pods["items"] if p["status"]["phase"] == "Running"]
```
rhcarvalho: I think my previous comment was unclear.

I meant:

```python
pods = json.loads(oc("-n", project, "get", "pods", "--show-all=false", "-o", "json"))["items"]
return [p["metadata"]["name"] for p in pods if p["status"]["phase"] == "Running"]
```

If we had tests, I possibly wouldn't have introduced this bug. I see the tests are empty for this function:
https://github.com/feedhenry/nagios-container/pull/7/files#diff-143e4678f975ff3a05e70c1a679a9fd9R7 :cry:

plugins/default/lib/openshift.py
```diff
@@ -5,18 +5,31 @@
...
+
 
 def get_running_pod_names(project):
-    return _get_running_pods(oc, project)
+    return _get_running_pod_names(oc, project)
```
rhcarvalho: Ha! Another bug of mine... if this was Go, the code wouldn't compile :)

plugins/default/lib/openshift.py
```diff
@@ -5,18 +5,31 @@
...
+
 
 def _exec_in_pods(oc, project, pods, cmd):
-    return [oc("-n", project, "exec", name, "--", *cmd) for name in pods]
+    return [[name, oc("-n", project, "exec", name, "--", *cmd)] for name in pods]
+
 
 def exec_in_pods(project, pods, cmd):
-    return _exec_in_all_runnning_pods(oc, project, pods, cmd)
+    return _exec_in_pods(oc, project, pods, cmd)
```
rhcarvalho: Ha! Another one! Thanks @jessesarn for fixing it.

plugins/default/lib/openshift.py
```diff
@@ -5,18 +5,31 @@
...
+
 
 def get_running_pod_names(project):
-    return _get_running_pods(oc, project)
+    return _get_running_pod_names(oc, project)
+
 
 def _exec_in_pods(oc, project, pods, cmd):
-    return [oc("-n", project, "exec", name, "--", *cmd) for name in pods]
+    return [[name, oc("-n", project, "exec", name, "--", *cmd)] for name in pods]
```
rhcarvalho: This is strange. We pass in the pod names as argument, normally we don't need the names in the output.
The problem here is where this function is called. We know the pod names there.

plugins/default/lib/disk_usage.py
rhcarvalho: In fact, we probably don't need any loop...

```python
# ...
pods = openshift.get_running_pod_names(project)
lines = openshift.exec_in_pods(project, pods, check_disk_cmd)
results = analize(pod, parse_df_lines(lines), warn, crit)
return report(results)
```

plugins/default/lib/disk_usage.py
rhcarvalho: Yeah, I never used `autopep8` myself, PEP8 is loose in several points, where "common sense" is suggested.
I suggested using the tool to give us some coding standard, and save some time in code review.

plugins/default/lib/disk_usage.py
rhcarvalho: How does `d` relate to all that you said? :)
`d` for dump?!

fhservices.cfg.j2
rhcarvalho: @david-martin it might be, looking at the JIRA alone, it's not clear there the documentation goes. If it is supposed to be a review of these notes, then we can review it when that JIRA gets worked on.

jessesarn: Yeah, these notes need to be updated to point to our documentation when we have it anyway, so I wouldn't put too much concern about the wording here until we have appropriate documentation.

plugins/default/lib/disk_usage.py
jessesarn: ok, I see what you mean, it felt weird giving names to a bunch of variables that I wasn't using, but I can see its certainly clearer.  Fixed.

plugins/default/lib/disk_usage.py
jessesarn: Yeah we do, when parsing the lines, we have multiple disk results lines per pod.  This is why I had asked if there was a good way to use map and pass in multiple parameters, but just using extend seems to be good enough to avoid the double looping.

plugins/default/lib/openshift.py
jessesarn: Oh, I see what you're saying.  Fixed.

plugins/default/lib/openshift.py
jessesarn: It was to make it easier to map the multi-line results back to the pod from which they came.  I guess I can use zip in the call to correlate the results back to the pod for which each belongs.  I.E.:
```
pods = openshift.get_running_pod_names(project)
    execs = openshift.exec_in_pods(project, pods, check_disk_cmd)
    for pod, lines in zip(pods, execs):
        results.extend(analize(pod, parse_df_lines(lines), warn, crit))
```

plugins/default/lib/disk_usage.py
rhcarvalho: It's "harder" / more code, because we are departing from function composition to explicitly for loops, more procedural code.

If you want to flatten all "lines" in a single list, you can use `sum`, for example:

```python
>>> lines = [[1, 1], [2, 2], [3, 3]]
>>> sum(lines, [])
[1, 1, 2, 2, 3, 3]
```

plugins/default/lib/openshift.py
rhcarvalho: You can use `zip`, but the problem is elsewhere...
Why does `analize` need to know about the pod name? It doesn't.
The analysis depend only on the thresholds and the percentages.

jessesarn:
@rhcarvalho One last review please... I hope.

plugins/default/lib/disk_usage.py
```diff
@@ -1,35 +1,95 @@
...
         return ()
-    return (mo.group(1), int(mo.group(2)), int(mo.group(3)))
+    return mo.group(3), int(mo.group(1)), int(mo.group(2))
+
+
+def parse_df_lines(lines):
+    return filter(None, map(parse_df_line, lines.splitlines()))
+
+
+def analize(pod, disks, warning_threshold, critical_threshold):
```
rhcarvalho: I haven't realized until looking further at the other changes.

Before, `analize` was focused on the "business logic": it takes just 4 numbers that it needs in order to tell if things are fine or not. It used to return two booleans answering two questions: "are we above warning level?", and "are we above critical level?".

Analyzing a container (a pod may have multiple containers, not sure if we're taking that into account) is a matter of analyzing with `analize` every line, and checking whether any of those booleans are true.

At this point I'm afraid I'm inflicting too much of a "personal preference" and sticking into the old code base into the review.

This PR adds a lot more code than I would expect, thinking that we were only missing the reporting part + bug fixes?

jessesarn: There are a few problems here:
1. A critical results shouldn't count as a warning as well, hence why I'm returning a nagios result code per check, something is either UNKNOWN, CRIT, WARN, or OK, in that order, it can't be more than one.
2. the original code didn't even run at all, and in particular the construct it suggested with map/analize didn't handle the fact that multiple rows were returned.  If you look at the changes to the main function, its 4 lines vs. 2 but handles the scenarios that weren't covered.
3. I can change it so analise just does the check part and doesn't take the pod name in but its going to get more confusing how the data is recombined in order to have the info it needs to report.

rhcarvalho:
@jessesarn let's care less about the form, and focus on working code. If it works for you I'm fine with merging.
The branch needs a rebase to resolve conflicts. I would also suggest squashing the commits.

jessesarn:
rebased, squashed commits, and fixed the complaints from travis

rhcarvalho:
Thanks @jessesarn , LGTM.

plugins/default/lib/disk_usage.py
```diff
@@ -44,5 +107,4 @@ def main():
     except:
         sys.exit(nagios.UNKNOWN)
```
rhcarvalho: @jessesarn comparing this with the component health check, I realized we're not exiting with the proper exit codes, it's either 0 or nagios.UNKNOWN.

I'll send a PR to address it.

PR #8, opened 2016-07-15
Rhmap 6722 add node core component checks

mikenairn:
* Adds checks for all the core node.js components
* Adds a new nagios module for querying the health endpoint (Copied from an existing check)

Some cross over with https://github.com/feedhenry/nagios-container/pull/5 regarding nagios plugins setup.

Note:

Currently the fh-appstore doesn't have the health endpoint so fails all the time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ EXPOSE 8080
 ADD sendEmail-epel-7.repo /etc/yum.repos.d/
 
 RUN yum install -y epel-release && \
-    INSTALL_PKGS="httpd nagios telnet supervisor python-jinja2 nagios-plugins-all sendEmail perl-Net-SSLeay perl-IO-Socket-SSL" && \
+    INSTALL_PKGS="httpd nagios telnet supervisor python-jinja2 nagios-plugins-all sendEmail perl-Net-SSLeay perl-IO-Socket-SSL jq" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN yum install -y epel-release && \
            -e 's|<Directory "/var/www">|<Directory "/usr/share/nagios/html">|' \
            /etc/httpd/conf/httpd.conf && \
     touch /supervisord.log /supervisord.pid && \
-    mkdir -p /var/log/nagios/archives /var/log/nagios/rw/ /var/log/nagios/spool/checkresults && \
+    mkdir -p /var/log/nagios/archives /var/log/nagios/rw/ /var/log/nagios/spool/checkresults /opt/rhmap/nagios/plugins/ && \
     chmod -R 777 /supervisord.log /supervisord.pid /var/log/nagios \
                  /etc/httpd /etc/passwd /var/log /etc/nagios /usr/lib64/nagios \
                  /var/spool/nagios /run /usr/share/httpd /usr/share/nagios && \
@@ -23,6 +23,8 @@ RUN yum install -y epel-release && \
 
 ADD supervisord.conf /etc/supervisord.conf
 ADD make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
+ADD plugins/default/* plugins/core/* /opt/rhmap/nagios/plugins/
+RUN chmod -R 755 /opt/rhmap/nagios/plugins/
 ADD start /start
 
 CMD ["/start"]
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -7,7 +7,7 @@ define command {
 # Check that a component responds to a /sys/info/health over HTTP
 define command {
        command_name   check_fh_component_http_health
-       command_line   $USER1$/check_http -H $ARG1$ -u /sys/info/health -p 8080
+       command_line   /opt/rhmap/nagios/plugins/fh-check-component-health.sh $ARG1$ 8080
 }
 
 define command {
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -15,14 +17,16 @@ RUN yum install -y epel-release && \
...
+    mkdir -p /var/log/nagios/archives /var/log/nagios/rw/ /var/log/nagios/spool/checkresults $NAGIOS_PLUGINS && \
     chmod -R 777 /supervisord.log /supervisord.pid /var/log/nagios \
                  /etc/httpd /etc/passwd /var/log /etc/nagios /usr/lib64/nagios \
                  /var/spool/nagios /run /usr/share/httpd /usr/share/nagios && \
     sed -i -e 's|cfg_file=/etc/nagios/objects/localhost.cfg||' /etc/nagios/nagios.cfg
 
 ADD supervisord.conf /etc/supervisord.conf
 ADD make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
+ADD /plugins/default/* /plugins/core/* $NAGIOS_PLUGINS
+RUN chmod -R 755 $NAGIOS_PLUGINS
```
rhcarvalho: @mikenairn here is a good example of where it may make sense to have another RUN instruction (as we have here) instead of combining all into one (as I mentioned elsewhere as a "best practice").
If we were to move this to the RUN instruction above, we'd have to move the ADD also.

Then the "problem" starts to be cache invalidation:

```Dockerfile
FROM ...

ADD /plugins/default/* /plugins/core/* $NAGIOS_PLUGINS
...
RUN yum install ... && \
        chmod -R 755 $NAGIOS_PLUGINS
```

That building that Dockerfile will re-run `yum install` at least as often as files in plugin/default change, and we normally want to avoid that to speed up local builds.

Release builds should be run without cache.

Dockerfile
```diff
@@ -15,14 +17,16 @@ RUN yum install -y epel-release && \
...
-    mkdir -p /var/log/nagios/archives /var/log/nagios/rw/ /var/log/nagios/spool/checkresults && \
+    mkdir -p /var/log/nagios/archives /var/log/nagios/rw/ /var/log/nagios/spool/checkresults $NAGIOS_PLUGINS && \
     chmod -R 777 /supervisord.log /supervisord.pid /var/log/nagios \
                  /etc/httpd /etc/passwd /var/log /etc/nagios /usr/lib64/nagios \
                  /var/spool/nagios /run /usr/share/httpd /usr/share/nagios && \
     sed -i -e 's|cfg_file=/etc/nagios/objects/localhost.cfg||' /etc/nagios/nagios.cfg
 
 ADD supervisord.conf /etc/supervisord.conf
 ADD make-nagios-fhservices-cfg make-nagios-commands-cfg fhservices.cfg.j2 commands.cfg.j2 /opt/rhmap/
+ADD /plugins/default/* /plugins/core/* $NAGIOS_PLUGINS
```
rhcarvalho: It's suspicious that the sources are absolute paths... can we double check the `/plugins`?

Dockerfile
```diff
@@ -4,8 +4,10 @@ EXPOSE 8080
 
 ADD sendEmail-epel-7.repo /etc/yum.repos.d/
 
+ENV NAGIOS_PLUGINS /opt/rhmap/nagios/plugins/
```
rhcarvalho: (optional)
I'd recommend using the syntax with an equal sign, that allows us to easily add more variables later.
https://docs.docker.com/engine/reference/builder/#/env

```Dockerfile
ENV NAGIOS_PLUGINS=/opt/rhmap/nagios/plugins/ \
    FOO=bar
```

rhcarvalho:
@mikenairn do we need the `NAGIOS_PLUGINS` variable  in the container at run time?

Dockerfile
mikenairn: I wondered about this also, we don't do this for anything else, would you expect this to be:

```
ADD plugins/default/* plugins/core/* $NAGIOS_PLUGINS
```

@jessesarn Any reason you made these absolute paths in your PR https://github.com/feedhenry/nagios-container/pull/5

jessesarn: Purely an accident.  We should make them relative.

mikenairn:
@rhcarvalho We probably don't tbh, we don't need it right now anyway. @jessesarn Do you think we need to have an env var for this?

jessesarn:
The reason I used an environment variable was for wrapping scripts.  For example, the mongodb check I created wraps an existing script to add the dynamic pod finding and to allow it to summarize the results of a check that only handles one endpoint.  Using the environment variable, we don't need to hard code the path inside the wrapper.  If we remove the environment variable, we'd have to make sure to sync up if we choose to relocate.

Dockerfile
rhcarvalho: Or really, if we don't need this env in the runtime, I'd just replace it's usages below.

We don't give env vars to things like `/var/log/nagios/spool/checkresults`, and other long paths, so having one for nagios plugins is at least inconsistent.

rhcarvalho:
@mikenairn @jessesarn we'll need to choose an approach and be consistent. The python scripts that process the jinja templates do write to hard coded paths.

Dockerfile
jessesarn: OK, we can take it out and I'll just modify my check to hard code the path.  As long as we all agree that /opt/rhmap/nagios/plugins is where we are going to put our scripts, then its no big deal.  I made that decision on my own, so having the environment variable made it easier to change if people had other opinions.

mikenairn:
@rhcarvalho Agreed. @jessesarn I'm going to remove the ENV instruction here and not add it as an env var as part of this PR since it's not required. You can update your PR as you see fit, but we should as @rhcarvalho points out be consistent in our approach.

Dockerfile
rhcarvalho: > /opt/rhmap/nagios/plugins

SGTM

rhcarvalho:
LGTM

PR #12, opened 2016-07-19
WIP disk check in Python

rhcarvalho:
The idea is to build some abstractions that can be reused in multiple
checks.

This is still not doing the same as what we've got with Bash in #7, but might be already something we can build on / discuss.

@mikenairn @pmccarthy @jessesarn (FYI @maleck13)


rhcarvalho:
Merged during daily standup to allow us to move forward based on some initial structure.

PR #13, opened 2016-07-20
Adding basic structure for Nagios Plugins

pmccarthy:
This purpose of this PR is to add a very basic directory structure for our Nagios python plugins. Once in place we can then add in some plugins and perhaps some examples or a better README.md. This is just a first pass.

@mikenairn @jessesarn @rhcarvalho

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -1,0 +1,13 @@
+#Nagios Plugin
+
+##Structure
+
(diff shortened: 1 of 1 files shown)


pmccarthy:
Closing this out as it is now longer required. See https://github.com/feedhenry/nagios-container/pull/12

PR #14, opened 2016-07-21
Add python based component health check

mikenairn:
Replaced bash check with new python check.

Jira:

https://issues.jboss.org/browse/RHMAP-9016


plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+        raise RequestError('The health endpoint at "%s" is not contactable. This could indicate that the service is down or the url is invalid!' % url)
+    return data
+
+def parse_response(response):
+    data = json.loads(response)
+    if data['status'] == 'ok':
+        nagios_status = nagios.OK
+    elif data['status'] == 'warn':
+        nagios_status = nagios.WARN
+    else:
```
jessesarn: For safety should be check for crit and return CRIT and then have the else fall back to UNKNOWN?

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python
+import nagios
```
rhcarvalho: Everyone has their favorite formula for organizing imports in Python. As a matter of organization, I tend to follow a formula that matches the Python's style guide, a.k.a. PEP8:

```python
import stdlib1
import stdlib2

import externallib1
import externallib2

import internallib1
import internallib2
```

The imports are organized into groups, with an empty line separating them. The first group are things from the Python standard library, then external packages/modules (we don't have any), then local pkgs/mods (the code we wrote ourselves).

This makes it easier to identify a file's dependency with a quick glance at the first lines.

The pattern is documented in https://www.python.org/dev/peps/pep-0008/#imports

@mikenairn wdyt?

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+        nagios_status = nagios.CRIT
+    return (nagios_status, data['summary'], data['details'])
+
+def report(summary, test_results):
+    print summary
+    for test in test_results:
+        print 'Test: %s - Status: %s - Details: %s' % (test['description'], test['test_status'],test['result'])
+
+def main():
+    parser = generate_parser().parse_args()
```
rhcarvalho: This is a bit misleading, the parser is what is returned by `generate_parser()`, while `parse_args()` return `args`?
https://docs.python.org/2/library/argparse.html#example

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+        sys.exit(nagios.CRIT)
+    nagios_status, test_summary, test_results = parse_response(response)
+    report(test_summary, test_results)
+    sys.exit(nagios_status)
+
+if __name__ == "__main__":
+    try:
+        main()
+    except CalledProcessError:
+        pass
```
rhcarvalho: These shouldn't be `pass`, it's just that I haven't finished it in my PR -- sorry!

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+        print error.message
+        sys.exit(nagios.CRIT)
+    nagios_status, test_summary, test_results = parse_response(response)
+    report(test_summary, test_results)
+    sys.exit(nagios_status)
+
+if __name__ == "__main__":
+    try:
+        main()
+    except CalledProcessError:
```
rhcarvalho: I suspect there is nothing in `main()` that can raise a `CalledProcessError`?
That exception is raised by functions from the `subprocess` module...

We can probably get rid of it here, and of the import.

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+        pass
+    except KeyError:
```
rhcarvalho: What do we expect to raise `KeyError`?

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+        pass
+    except Exception:
```
rhcarvalho: `except:` and `except Exception:` are deceptively not the same :( -- thank you, Python!

https://wiki.python.org/moin/HandlingExceptions#General_Error_Handling

In Python 2, there are exceptions that are not subclasses of `Exception` nor `BaseException`. E.g., any old-style class (one that doesn't inherit from `object`) can be raised as well:

```console
$ python
Python 2.7.11 (default, Jul 10 2016, 20:58:20) 
[GCC 6.1.1 20160621 (Red Hat 6.1.1-3)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> try:
...     class FooErr:
...         pass
...     raise FooErr
... except Exception as e:                                                                 
...     print "got it:", e
... except:
...     print "not caught"
... 
not caught
```

We don't raise those exceptions ourselves, but it's hard to tell whether code we call doesn't do so. Maybe all the code in the stdlib at least inherits from `BaseException`, but I am not sure about that.

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+        print("Unexpected error:", sys.exc_info()[0])
```
rhcarvalho: It is inconsistent to use parentheses in a call to print and don't use it in other places. 

In Python 2, `print` is a statement, it takes no parentheses.
In Python 3, `print` is a function. It is possible to make `print` be a function in Python via the magic import:

```python
from __future__ import print_function
```

But I don't see any benefit for us in doing that for these nagios checks.

I'm in favor of removing the parentheses, consistently embracing the Python 2 style.

plugins/default/lib/test_component_health.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
+import unittest
+
+from component_health import parse_response
+
+class TestParseResponse(unittest.TestCase):
+    def runTest(self):
+        self.assertEqual(parse_response("{\"status\": \"ok\", \"summary\": \"Test Summary\", \"details\": []}"), (0, 'Test Summary', []))
+        self.assertEqual(parse_response("{\"status\": \"warn\", \"summary\": \"Test Summary\", \"details\": []}"), (1, 'Test Summary', []))
+        self.assertEqual(parse_response("{\"status\": \"crit\", \"summary\": \"Test Summary\", \"details\": []}"), (2, 'Test Summary', []))
```
rhcarvalho: Maybe instead of the magic 0, 1, 2, here we really want to use the nagios constants?

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,76 @@
...
+            # For some reason the health endpoint returns a http status code of 500 when the health isn't ok, so we have to check the http error for a health response here
+            try:
+                data = error.read()
+                json.loads(data)
+            except ValueError:
+                raise RequestError(error)
+        else:
+            raise RequestError(error)
+    except urllib2.URLError:
+        raise RequestError('The health endpoint at "%s" is not contactable. This could indicate that the service is down or the url is invalid!' % url)
```
rhcarvalho: pro tip: always use explicit tuples when string formatting with `%`, otherwise funny things may happen.

```python
'The health endpoint at "%s" is ...' % (url,)
```

plugins/default/lib/component_health.py
mikenairn: The problem i have with not explicitly saying 'Excpetion' was that it was catching the sys.exit as well:

```
[vagrant@ose32 lib]$ python component_health.py -H 172.30.238.128 -P 8080 -E /box/api/health
millicore fully operational
Test: Try to call ping endpoint of memcache - Status: ok - Details: memcache fully operational
Test: Try to call ping endpoint of mysql - Status: ok - Details: mysql fully operational
Test: Try to call ping endpoint of fh-supercore - Status: ok - Details: fh-supercore fully operational
Test: Try to call ping endpoint of fh-aaa - Status: ok - Details: fh-aaa fully operational
Test: Try to call ping endpoint of fh-scm - Status: ok - Details: fh-scm fully operational
('Unexpected error:', <type 'exceptions.SystemExit'>)
[vagrant@ose32 lib]$ echo $?
3
```
The return codes are important in nagios to get the correct status, but this way its making every status 3 (UNKNOWN).

What do you recommend here?

plugins/default/lib/test_component_health.py
mikenairn: The exit status codes are important for nagios, so I would rather test for what we want explicitly in the test. We could of course use the nagios constants, but then you can change those values and not have the tests fail which I would prefer didn't happen.

plugins/default/lib/component_health.py
mikenairn: Fine by me, if there is a recommended way of organising these, lets do that.

plugins/default/lib/test_component_health.py
rhcarvalho: It they need to match nagios, the constants should not change. That's not something we can control with these tests.

At the same time/PR one would change nagios.CRIT to 42, he/she would also fix the test by changing the hard coded value to 42. So not using the constant doesn't fix the problem, don't you think?

plugins/default/lib/component_health.py
rhcarvalho: I see... one "workaround" would be to explicitly catch `SystemExit`, but that would be silly :)

Probably a better design starts with `main` never calling `sys.exit`. Instead we should make it return the exit code. Something like:

```python
if __name__ == "__main__":
    try:
        nagios_code = main()
    except ...:
        ....
    except:
        ....
    sys.exit(nagios_code)
```

This way `main` would also become testable, you can verify that it returns the right code without terminating the program (and the test runner).

rhcarvalho:
@mikenairn @jessesarn with this we can probably get rid of epel and jq from the Dockerfile?

rhcarvalho:
How about we use something like https://pypi.python.org/pypi/autopep8 to format our Python files?

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,74 @@
...
+def parse_response(response):
+    data = json.loads(response)
+    if data['status'] == 'ok':
+        nagios_status = nagios.OK
+    elif data['status'] == 'warn':
+        nagios_status = nagios.WARN
+    elif data['status'] == 'crit':
+        nagios_status = nagios.CRIT
+    else:
+        nagios_status = nagios.UNKNOWN
```
rhcarvalho: The existing code is perfectly fine, just sharing a Python idiom with a dict in case it comes in hand some day:

```python
def parse_response_alt(response):
    status = {
      "ok": nagios.OK,
      "warn": nagios.WARN,
      "crit": nagios.CRIT
    }.get(data["status"], nagios.UNKOWN)
    return status, data.get("summary", ""), data.get("details", "")
```

Using `dict.get` vs. `dict[key]` has a subtle difference in that `parse_response` may raise a KeyError for bad input, while `parse_response_alt` doesn't, probably not relevant here.

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,74 @@
...
+
+def do_request(host, port, endpoint):
+    url =  'http://%s:%s%s' % (host,port,endpoint)
+    try:
+        data = urllib2.urlopen(url).read()
+    except urllib2.HTTPError, error:
+        if error.code == 500:
+            # For some reason the health endpoint returns a http status code of 500 when the health isn't ok, so we have to check the http error for a health response here
+            try:
+                data = error.read()
```
rhcarvalho: The described behavior of the health endpoint is funny, but I trust it. However, I couldn't understand how are we reading data from `error`?

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,74 @@
...
+    parser.add_argument("-H","--host", action="store", required=True, help="Host")
+    parser.add_argument('-P','--port', action="store", required=True, help="Port")
+    parser.add_argument("-E","--endpoint", action="store", required=False, help="Endpoint", default="/sys/info/health")
+    return parser
+
+def do_request(host, port, endpoint):
+    url =  'http://%s:%s%s' % (host,port,endpoint)
+    try:
+        data = urllib2.urlopen(url).read()
+    except urllib2.HTTPError, error:
```
rhcarvalho: In "modern Python", the `except klass as e:` syntax is normally preferred over the `,`, for that old syntax is often associated with mistakes when we list multiple exception classes like `except (KeyError, ValueError):`.

https://docs.python.org/2/tutorial/errors.html#handling-exceptions

plugins/default/lib/component_health.py
mikenairn: The health endpoint returns a http status of 500 if the health is deemed to be anything other than 'ok', even though the http request is perfectly fine and there is no internal server error at the application level. It seems to be a design thing to use the status codes in this way. I wouldn't agree with it, just makes things harder,  but that's the way it is.

The urllib2 error contains the json response (health response) in the error see http://stackoverflow.com/questions/2233687/overriding-urllib2-httperror-and-reading-response-html-anyway

plugins/default/lib/component_health.py
mikenairn: I had something similar to this at one point, and then didn't like the fact that main was returning a value for some reason and thought it was thing that should be dealing with the sys.exit.

rhcarvalho: I can understand, perhaps the name `main` make us think differently.

philbrookes:
@rhcarvalho as we mentioned on bluejeans, it'd be a nice idea to have the number of lines taken from the logfiles as an optional argument to the dump tool.

rhcarvalho:
@philbrookes wrong PR, but I'll add a checkbox to https://github.com/fheng/fh-system-dump-tool/pull/4

plugins/default/lib/test_component_health.py
mikenairn: I added a separate test for the nagios lib, which makes me happier to use it here.

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,101 @@
...
+def main():
+    try:
+        args = generate_parser().parse_args()
+        host = args.host
+        port = args.port
+        endpoint = args.endpoint
+        response = do_request(host, port, endpoint)
+        nagios_status, test_summary, test_results = parse_response(response)
+        report(test_summary, test_results)
+        sys.exit(nagios_status)
```
mikenairn: This line needs removed

plugins/default/lib/test_nagios.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
+import unittest
+
+import nagios
+
+
+class TestParseResponse(unittest.TestCase):
```
rhcarvalho: The name here is wrong, copy pasted from the other file :)

I don't see the point of testing constants.

mikenairn: Normally I might agree, but these are very important to **nagios**. To me if something is being tested it means it's important and might make you think twice if you go changing it.

plugins/default/lib/component_health.py
```diff
@@ -0,0 +1,100 @@
...
+            try:
+                data = error.read()
+                json.loads(data)
+            except ValueError:
+                raise RequestError(error)
+        else:
+            raise RequestError(error)
+    except urllib2.URLError:
+        raise RequestError(
+            'The health endpoint at "%s" is not contactable. This could indicate that the service is down or the url is invalid!' %
```
rhcarvalho: Shouldn't we include here the original error?

I would rather see a shorter message + the cause of the failure.

rhcarvalho:
@mikenairn I think we can improve the message in case of an HTTP error to include the failure reason and remove `jq` (and epel?) from the Dockerfile and LGTM.

plugins/default/lib/component_health.py
mikenairn: Possibly. Thing to note is that whatever this returns will be printed in the status of the nagios check in the dashboard, and just having a http error might not give you much of a clue unless you are familiar with the check itself.

mikenairn:
@rhcarvalho Removed the jq package, epel repo is still required for sendEmail package. Also updated that error to include the http response. Merging this now.

plugins/default/lib/component_health.py
rhcarvalho: Shouldn't be just the error, but a user friendly message + details.

Imagine the error is a connection timeout, or DNS error. It would be useful to be able to differentiate that from the health endpoint reporting a problem.

If we omit the original error, we can't determine what actually happened.

PR #15, opened 2016-07-22
Adding tcp check for memcached

pmccarthy:
As check uses the existing nagios check_tcp plugin, there was no need for additional python plugins. Only change required was a new command and service in the fhservices.cfg.j2 file

@jessesarn @mikenairn @rhcarvalho mind taking a look?

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -10,6 +10,12 @@ define command {
        command_line   /opt/rhmap/nagios/plugins/fh-check-component-health.sh $ARG1$ 8080
 }
 
+# Check for valid response from Memcached
+define command {
+       command_name   check_memcached
+       command_line   $USER1$/check_tcp -H memcached -p $ARG1$ -E -s 'version\n' -e 'VERSION' -w2 -c5 -t5
+}
+
 define command {
        command_name   check_nagios_host
        command_line   $USER1$/check_http -H $HOSTADDRESS$ -S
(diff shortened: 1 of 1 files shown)


fhservices.cfg.j2
```diff
@@ -10,6 +10,12 @@ define command {
        command_line   /opt/rhmap/nagios/plugins/fh-check-component-health.sh $ARG1$ 8080
 }
 
+# Check for valid response from Memcached
+define command {
+       command_name   check_memcached
+       command_line   $USER1$/check_tcp -H memcached -p $ARG1$ -E -s 'version\n' -e 'VERSION' -w2 -c5 -t5
```
rhcarvalho: What are those version arguments?

What do the thresholds mean? 2 and 5.

fhservices.cfg.j2
```diff
@@ -74,3 +80,13 @@ define service {
...
 }
 {% endfor %}
+
+define service {
+      service_description memcached:ping
+      hosts {{ rhmap_router_dns }}
+      check_command check_memcached!11211
+      use generic-service
+
+      notes This should should be able to communicate with memcached on port 11211 and cannot. Check it is running and this server can communicate with it on that port.
```
rhcarvalho: s/should should/should

rhcarvalho:
Looks reasonable.

fhservices.cfg.j2
pmccarthy: @rhcarvalho In our SaaS grids there were some issues around the version of memcached being used so by passing in these version arguments we're getting the version being used, may not be relevant for On-Prem but dont see any harm in leaving it there.

The thresholds are:
-w <warning-response-time>
-c <critical-response-time>
-t <timeout>

Each of these are measured in seconds.

fhservices.cfg.j2
pmccarthy: @rhcarvalho Well spotted, fixing this typo now

fhservices.cfg.j2
rhcarvalho: Why is critical == timeout?

The explanation gives the general idea, thanks, but what does `-E -s 'version\n' -e 'VERSION'` do?

PR #16, opened 2016-07-22
Rhmap 8088 add millicore check

mikenairn:
* Adds a new health check for millicore 	7f57c2c
* Modifies the way the services are generated to allow more reuse  4c0d231

Requires https://github.com/feedhenry/nagios-container/pull/14 to be merged first.

Jira:

https://issues.jboss.org/browse/RHMAP-8088

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -6,8 +6,8 @@ define command {
 
 # Check that a component responds to a /sys/info/health over HTTP
 define command {
-       command_name   check_fh_component_http_health
-       command_line   /opt/rhmap/nagios/plugins/fh-check-component-health -H $ARG1$ -P 8080
+       command_name   check_fh_component_health
+       command_line   /opt/rhmap/nagios/plugins/fh-check-component-health -H $ARG1$ -P $ARG2$ -E $ARG3$
 }
 
 # Check for valid response from Memcached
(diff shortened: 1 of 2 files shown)


fhservices.cfg.j2
```diff
@@ -57,26 +57,26 @@ define host {
...
 {% endfor %}
 
-{% for fhservice in fh_services_health %}
+{% for fhservice in fh_services if 'health' in fhservice['checks'] %}
 define service {
-       service_description {{ fhservice }}::Health
+       service_description {{ fhservice['name'] }}::Health
        hosts {{ rhmap_router_dns }}
-       check_command check_fh_component_http_health!{{ fhservice }}
+       check_command check_fh_component_health!{{ fhservice['name'] }}!{{ fhservice['port'] if 'port' in fhservice  else '8080' }}!{{ fhservice['health_endpoint'] if 'health_endpoint' in fhservice  else '/sys/info/health' }}
```
rhcarvalho: I'm not sure what is valid in the jinja template, but perhaps if supports calling dict methods?

```python
fhservice.get("port", "8080")
fhservice.get("health_endpoint", "/sys/info/health")
```

rhcarvalho:
@mikenairn IIUC this is not only adding checks for millicore but several other components too? Nice! Could you update the title/description to reflect that?

fhservices.cfg.j2
mikenairn: Modified this to use the dict.get() method, looks a lot cleaner and works fine in the jinja template.

mikenairn:
@rhcarvalho This only adds 1 new check for millicore https://github.com/feedhenry/nagios-container/pull/16/commits/29ddaa4df0fbabd71d1bd15835915475fc91f13f, the other checks already existed. This commit https://github.com/feedhenry/nagios-container/pull/16/commits/7df921d72872cab2150e678e320c434469cd67cd modified the config being passed in to combine everything and to make it more configurable.

fhservices.cfg.j2
```diff
@@ -57,26 +57,26 @@ define host {
        host_name {{ rhmap_router_dns }}
 }
 
-{% for fhservice in fh_services_ping %}
+{% for fhservice in fh_services if 'ping' in fhservice.get("checks") %}
```
rhcarvalho: `dict.get` with a single arguments means the default is `None`. It's a `TypeError` to say `if 'ping' in None`.

Are we implicitly treating some dictionary keys as "required"?

rhcarvalho:
> This only adds 1 new check for millicore...

That's right, got it. Sorry I failed to notice that on the first pass.

fhservices.cfg.j2
mikenairn: There is no obvious sensible default for 'name' or 'checks' in a service config. We could have that default to an empty array, but that would only mask a potential issues with the config being passed in. Currently if this fails it will create an invalid config and the container will fail to start which seems preferable. wdyt?

rhcarvalho: Where is the code that processes the template?
Is the template processed in run time or build time?

rhcarvalho: > wdyt?

I think if I was maintaining this and came across a problem, I'd prefer to see a KeyError rather than maybe a TypeError.
In the case where we do `... in ...`, we'd get a TypeError, but in other places it would just produce the string "None" in the nagios config, not so easy to catch.

mikenairn: It's processed at run time here https://github.com/feedhenry/nagios-container/blob/master/start#L14

rhcarvalho: Alright, so the `/start` script should fail and the container keep restarting.
For the ones that have no default value, can we use the direct key access as we had before? That would help us detect eventual problems earlier.

mikenairn: Yeah we can revert those ones back to accessing the keys directly

philbrookes:
:+1: Verified on my local core:
1) updated nagios image and verified test was green.
2) destroyed the millicore pod and the test went red.
3) millicore pod came up and test went back to green.

rhcarvalho:
LGTM

PR #17, opened 2016-07-25
RHMAP-8089 - Add UPS health check

mikenairn:
* Add UPS health check

Jira:

https://issues.jboss.org/browse/RHMAP-8089


pmccarthy:
@mikenairn Ran this against my local Core and can confirm that Nagios is showing green for UPS

PR #18, opened 2016-07-25
Rhmap 9094 add travis file

mikenairn:
* Adds a travis build file, runs pep8 and unit tests
* Executed autopep8 on all existing plugins (no functional changes)
* Add build status to readme

Jira:

https://issues.jboss.org/browse/RHMAP-9094

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,39 @@
 # RHMAP Nagios for Docker
 
+[![Build Status](https://travis-ci.org/feedhenry/nagios-container.svg?branch=master)](https://travis-ci.org/feedhenry/nagios-container)
+
 Nagios server in a Docker container for RHMAP.
+
+## Development
+
+Install dependencies:
+```
+pip install pep8
+pip install autopep8
+```
+
+Run tests:
+
+```
+python -m unittest discover -s plugins/default
+````
+
+Style Guide:
+
+All python modules should conform to the PEP8 style guide https://www.python.org/dev/peps/pep-0008/
+
+To verify:
+
+```
+pep8 plugins/ --show-source
+```
+
+More info https://pypi.python.org/pypi/pep8
+
+To format code:
+
+```
+autopep8 --in-place --aggressive --aggressive plugins/default/**/*.py
+```
+
+More info https://pypi.python.org/pypi/autopep8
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1,3 +1,5 @@
 # RHMAP Nagios for Docker
 
+[![Build Status](https://travis-ci.org/feedhenry/nagios-container.svg?branch=master)](https://travis-ci.org/feedhenry/nagios-container)
```
pmccarthy: Is it also worth adding in a section which outlines the two validation commands being run as part of the travis build? i.e. pep8 plugins/ --ignore E501 --show-source and python -m unittest discover -s plugins/default so that people know how to run this tests locally beforehand.

mikenairn: Yes it would, let me update this.

pmccarthy:
Overall this looks fine. Will be nice to have a style guide in place to ensure consistency.

mikenairn:
@pmccarthy Added some test and style verification instructions to the readme caddd23b3bbef2f43e989a41674c5d660e35785a

README.md
```diff
@@ -1,3 +1,39 @@
...
+````
+
+Style Guide:
+
+All python modules should conform to the PEP8 style guide https://www.python.org/dev/peps/pep-0008/
+
+To verify:
+
+```
+pep8 plugins/ --show-source
```
pmccarthy: @mikenairn  Ran through the commands in this guide, for the pep8 command, a bunch of E501 errors were displayed for me locally. I know these are ignored as part of the travis build itself, but could be confusing to anyone that is unaware of this. Maybe a note about E501 would be helpful

mikenairn: Updated to make the command the same. The  E501 error is about line lengths, you can make it allow longer lines but I think its safe enough to ignore them all for now.

mikenairn:
@pmccarthy Change of plan, I added a setup.cfg and modified the allowed line length for pep8 checks. This config gets shared by both pep8 and autopep8 so seems the best way to configure it.

pmccarthy:
@mikenairn Went through the readme again, can confirm that all commands are working as expected now, no more E501 errors which verifies the setup.cfg property is working. Also ran this version of the nagios container and can confirm that all checks are passing as expected.

PR #19, opened 2016-07-26
Adding keycloak health endpoint check

pmccarthy:
Adding custom health check for Keycloak. Not adding this to the standard block of health checks as the Keycloak endpoint has an expected HTTP 302 redirect response which needs to be handled by the standard check_http nagios plugin.

Related Jira:
https://issues.jboss.org/browse/RHMAP-8091

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -10,6 +10,12 @@ define command {
        command_line   /opt/rhmap/nagios/plugins/fh-check-component-health -H $ARG1$ -P $ARG2$ -E $ARG3$
 }
 
+# Check that an EAP component responds to a ping over HTTP
+define command {
+       command_name   check_eap_component_health
+       command_line   $USER1$/check_http -H $ARG1$ -p $ARG2$ -u $ARG3$ -e "$ARG4$"
+}
+
 # Check for valid response from Memcached
 define command {
        command_name   check_memcached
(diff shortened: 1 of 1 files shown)


fhservices.cfg.j2
```diff
@@ -82,11 +88,21 @@ define service {
 {% endfor %}
 
 define service {
+      service_description keycloak:Health
+      hosts {{ rhmap_router_dns }}
```
mikenairn: This is fine, but depending on whether this gets merged before https://github.com/feedhenry/nagios-container/pull/20 or not, one of them will need to change this to be:

```
define service {
      service_description keycloak:Health
      check_command check_eap_component_health!ups!8080!/auth/admin!HTTP/1.1 302
      use generic-service
      hostgroup_name core
      notes This server failed to get a successful response from the Keycloak service health endpoint. Check the response code & body and ensure the service and its dependencies are running and configured correctly.
      contact_groups rhmapadmins
}
```

pmccarthy: @mikenairn Probably best to get https://github.com/feedhenry/nagios-container/pull/20 merged first in this case. Only a very minor change in this PR then to make it work. I can look at that PR now.

mikenairn:
This Look good to me. Verified locally.

pmccarthy:
Merging this now, @mikenairn is then going to update https://github.com/feedhenry/nagios-container/pull/20 to fit this change in.

PR #20, opened 2016-07-26
Rhmap 8605 use hostgroups for core or mbaas

mikenairn:
* Allows the nagios image to be shared between both the rhmap core and mbaas.
* Uses an env var to switch between the 2 or both (RHMAP_HOSTGROUPS='core'|'mbaas'|'core,mbaas')
* Uses nagios hostgroups to determine what checks should be applied to what, all services must have a hostgroup_name set on them.

Jira:

https://issues.jboss.org/browse/RHMAP-8605

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -71,7 +71,7 @@ define host {
        use fh-component-container
        address {{ rhmap_router_dns }}
        host_name {{ rhmap_router_dns }}
-       hostgroups core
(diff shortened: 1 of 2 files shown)


david-martin:
Reviewed and tested locally. Results look as expected @mikenairn if you want to take a look at the comment in https://issues.jboss.org/browse/RHMAP-8605

mikenairn:
@david-martin Results look as expected, this looks ready to merge now.

PR #21, opened 2016-07-26
RHMAP-9106 - Add mbaas specific checks based on https://github.com/fe…

mikenairn:
* Adds mbaas component health and ping checks based on what is currently configured here https://github.com/feedhenry/mbaas-monitoring/blob/master/make-nagios-fhservices-cfg.py#L7-L8

Jira:

https://issues.jboss.org/browse/RHMAP-9106


david-martin:
@mikenairn Changes look trivial enough.
I'll try this out locally now

david-martin:
@mikenairn tested locally with expected results (see jira)

PR #22, opened 2016-07-31
Use flake8 instead of pep8

rhcarvalho:
- Run `flake8` instead of `pep8` on all Python code in the repository
- Suggest auto-formatting all files
- Update README.md


jessesarn:
👍  Looks good

PR #23, opened 2016-08-01
Uniform checks

rhcarvalho:
This is making the structure of the existing checks more similar to facilitate adding new checks in the same style.

Each script looks like:

```python
#!/usr/bin/env python
import ...
import sys
import traceback

import nagios


def generate_parser():
    parser = argparse.ArgumentParser(
        description="...",
    )
    parser.add_argument(
        # ...
    )
    # ...
    return parser


# ...


def check(...):
    # ...
    # return nagios codes
    return nagios.OK


if __name__ == "__main__":
    args = generate_parser().parse_args()
    code = nagios.UNKNOWN
    try:
        code = check(...)
    except:
        traceback.print_exc()
    finally:
        sys.exit(code)
```

## Functional changes

The functional changes squeezed in here:

- The `fh-check-component-health` check returns `UNKNOWN` instead of `CRIT` for connection errors
- Fixed exit codes for `disk-usage` check:  it only ever returned 0 (implicit when the script terminates) or UNKNOWN in case of errors other than CalledProcessError and KeyError (both incorrectly returned 0)

### DIFF
--- a/plugins/default/lib/component_health.py
+++ b/plugins/default/lib/component_health.py
@@ -76,13 +76,9 @@ def report(summary, test_results):
 
 
 def check(host, port, endpoint):
-    try:
-        response = do_request(host, port, endpoint)
-        nagios_status, test_summary, test_results = parse_response(response)
-        report(test_summary, test_results)
-    except RequestError as error:
-        print error.message
-        nagios_status = nagios.CRIT
+    response = do_request(host, port, endpoint)
+    nagios_status, test_summary, test_results = parse_response(response)
+    report(test_summary, test_results)
     return nagios_status
 
 
--- a/plugins/default/lib/component_health.py
+++ b/plugins/default/lib/component_health.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
-import sys
-import json
 import argparse
+import json
+import sys
+import traceback
 import urllib2
 
 import nagios
@@ -74,22 +75,23 @@ def report(summary, test_results):
         print 'Test: %s - Status: %s - Details: %s' % (test['description'], test['test_status'], test['result'])
 
 
-def main():
-    args = generate_parser().parse_args()
+def check(host, port, endpoint):
     try:
-        host = args.host
-        port = args.port
-        endpoint = args.endpoint
         response = do_request(host, port, endpoint)
         nagios_status, test_summary, test_results = parse_response(response)
         report(test_summary, test_results)
     except RequestError as error:
         print error.message
         nagios_status = nagios.CRIT
-    except:
-        print "Unexpected error: %s" % (sys.exc_info()[0])
-        nagios_status = nagios.UNKNOWN
-    sys.exit(nagios_status)
+    return nagios_status
+
 
 if __name__ == "__main__":
-    main()
+    args = generate_parser().parse_args()
+    code = nagios.UNKNOWN
+    try:
+        code = check(args.host, args.port, args.endpoint)
+    except:
+        traceback.print_exc()
+    finally:
+        sys.exit(code)
--- a/plugins/default/lib/component_health.py
+++ b/plugins/default/lib/component_health.py
@@ -7,34 +7,29 @@ import urllib2
 import nagios
 
 
+def generate_parser():
+    parser = argparse.ArgumentParser(
+        description="Checks the health endpoint of an RHMAP component",
+    )
+    parser.add_argument(
+        "-H", "--host", required=True,
+        help="host name of the component",
+    )
+    parser.add_argument(
+        "-P", "--port", default="8080",
+        help="port number of the component (default: %(default)s)",
+    )
+    parser.add_argument(
+        "-E", "--endpoint", default="/sys/info/health",
+        help="health endpoint (default: %(default)s)",
+    )
+    return parser
+
+
 class RequestError(Exception):
 
     def __init__(self, message):
         self.message = message
-
-
-def generate_parser():
-    parser = argparse.ArgumentParser()
-    parser.add_argument(
-        "-H",
-        "--host",
-        action="store",
-        required=True,
-        help="Host")
-    parser.add_argument(
-        '-P',
-        '--port',
-        action="store",
-        required=True,
-        help="Port")
-    parser.add_argument(
-        "-E",
-        "--endpoint",
-        action="store",
-        required=False,
-        help="Endpoint",
-        default="/sys/info/health")
-    return parser
 
 
 def do_request(host, port, endpoint):
--- a/plugins/default/lib/component_health.py
+++ b/plugins/default/lib/component_health.py
@@ -32,6 +32,9 @@ class RequestError(Exception):
     def __init__(self, message):
         self.message = message
 
+    def __str__(self):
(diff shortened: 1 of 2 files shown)


fhservices.cfg.j2
```diff
@@ -121,6 +121,7 @@ define service {
 
 define service {
        service_description Storage::Pod::Disk Usage
+       # REVIEW: is this being applied to every pod/container? How does it work?
```
rhcarvalho: @jessesarn could you help me understand this?

plugins/default/lib/component_health.py
```diff
@@ -79,22 +75,25 @@ def report(summary, test_results):
...
-        host = args.host
-        port = args.port
-        endpoint = args.endpoint
         response = do_request(host, port, endpoint)
         nagios_status, test_summary, test_results = parse_response(response)
         report(test_summary, test_results)
     except RequestError as error:
         print error.message
+        # REVIEW: shouldn't this be nagios.UNKNOWN? If the request failed, we
+        # "don't know" if the component is healthy or not.
```
rhcarvalho: @jessesarn @mikenairn @pmccarthy opinions here?

@jessesarn in terms of UI, what is the effect of getting an UNKNOWN versus CRITICAL?

rhcarvalho: By UI I'd also think of email notifications, not only web interface.

plugins/default/lib/disk_usage.py
```diff
@@ -53,6 +62,10 @@ def analize(pod, disks, warning_threshold, critical_threshold):
 
 
 def report(results):
+    # REVIEW: is nagios.UNKNOWN the best here?
```
rhcarvalho: This is handling the case when we didn't get any lines from `df`.

@jessesarn what would you suggest we do in this case?

fhservices.cfg.j2
jessesarn: At the nagios config level, only one service is being defined.  That service runs a check which dynamically finds all the pods to run the check command to summarize the results.  The overall status returned to nagios is based on the "worst" status of underlying checks (order UNKNOWN, CRIT, WARN, OK).

plugins/default/lib/disk_usage.py
jessesarn: As we are no longer filtering out tmpfs mounts, if we aren't getting a result from df, something is seriously wrong.  The overall status reported to nagios is taking the highest numeric value, so just one missing df result will show as UNKNOWN.  I honestly can't forsee a situation where we would get one empty df result yet get all the rest just fine.

fhservices.cfg.j2
rhcarvalho: What exactly does the pod discovery? Could you send a link to it?
We need to ensure this works for multi-container pods.

plugins/default/lib/component_health.py
rhcarvalho: > Since we have full control of all of the containers/services in our RHMAP projects, every check should be capable of determining an exact status without encountering UNKNOWN.

This is the case when there was a request error -- e.g. the service returned something unexpected.

For now, see the line below, we're returning `nagios.CRIT` in that case, but I suspect it would be more appropriate to say `nagios.UNKNOWN`, because a request error means there was an error performing the check.

Do you agree?

plugins/default/lib/disk_usage.py
rhcarvalho: Do you agree with `nagios.UNKNOWN` (and not `nagios.CRIT`, or others)?

jessesarn: I think nagios.UNKNOWN makes the most sense, but I am not against changing it if somebody has a good argument otherwise.

plugins/default/lib/component_health.py
rhcarvalho: Talked with @mikenairn and @philbrookes on IRC, there's agreement on making this `nagios.UNKNOWN`.

However, we'll need to revisit the ping check.

fhservices.cfg.j2
rhcarvalho: I hope whatever we do is not relying on IP addresses:

> "Kubernetes applies IP addresses at the Pod scope - containers within a Pod share their network namespaces - including their IP address."

http://kubernetes.io/docs/admin/networking/#kubernetes-model


It seems that the disk check runs against the local file system, still puzzled how it is run for all containers / all disks.

rhcarvalho: Thanks to @mikenairn I got a clear picture. I was missing the call to `oc exec` on all pods...

https://github.com/feedhenry/nagios-container/blob/1bc02dbd22db2741bb726944d95b43f9e796258b/plugins/default/lib/disk_usage.py#L96

The problem then is that that call doesn't handle pods with multiple containers.

fhservices.cfg.j2
```diff
@@ -121,6 +121,7 @@ define service {
...
        hosts {{ rhmap_router_dns }}
```
mikenairn: @jessesarn I see this was added in a previous PR, but just wondered should we be using 'hostgroup_name' here instead of 'hosts' like all the other services?

```
hostgroup_name core,mbaas
```

It will work like this since it's supposed to be applied in all cases, but might be best to keep them all consistent

jessesarn: Yeah, good call.  Will fix.

rhcarvalho:
@jhellar where you able to verify this?

jhellar:
@rhcarvalho yes. LGTM.

rhcarvalho:
@jhellar thanks

plugins/default/lib/component_health.py
```diff
@@ -79,22 +75,19 @@ def report(summary, test_results):
...
 if __name__ == "__main__":
-    main()
+    args = generate_parser().parse_args()
+    code = nagios.UNKNOWN
+    try:
+        code = check(args.host, args.port, args.endpoint)
+    except:
+        traceback.print_exc()
+    finally:
+        sys.exit(code)
```
mikenairn: This is now showing a stacktrace and a not very useful message in an error case for a status in the UI of nagios.

```
Status Information:	(No output on stdout) stderr: Traceback (most recent call last):
File "/opt/rhmap/nagios/plugins/fh-check-component-health", line 89, in <module>
code = check(args.host, args.port, args.endpoint)
File "/opt/rhmap/nagios/plugins/fh-check-component-health", line 79, in check
response = do_request(host, port, endpoint)
File "/opt/rhmap/nagios/plugins/fh-check-component-health", line 55, in do_request
(url, error))
RequestError
```

Previously it used to just be:

```
connect to address fh-appstore and port 8080: No route to host
```

rhcarvalho: I see, that's because our `RequestError` class doesn't implement the `__str__` method... so that `message` attribute is ignored.

rhcarvalho: Would you be ok with the traceback + the message?
Or just the message?

rhcarvalho:
Anyone, feel free to merge this -- I'm on PTO till Tue.

PR #25, opened 2016-08-11
add support for multiple containers

jessesarn:


plugins/default/lib/openshift.py
```diff
@@ -44,7 +44,51 @@ def exec_in_pods(project, pods, cmd):
...
+    #         for p in pods if p["status"]["phase"] == "Running"]
+    # ))
+
+
+def get_running_pod_containers(project, selector=None):
+    return _get_running_pod_containers(oc, project, selector)
+
+
 def get_project():
+    # return oc("project", "-q").rstrip("\n")
```
philbrookes: Shall we just delete commented code?

jessesarn: Yeah, I keep putting that back in so I can easily test the python from my mac without having to put code back into the pod.  I'll remove it before merging.

jessesarn:
@philbrookes Also, I'll fix the flake8 complaints on the test script before merging.

philbrookes:
:+1: Code looks fine, and verified on my local ose32

vchepeli:
LGTM, Motivation and Changes are not presented in PR description!

PR #27, opened 2016-08-17
Adding support for mbaas 3 node template

pmccarthy:
This change is required to support the existing 3 node mbaas template where the service names are different to what we have in our core templates, see related Jira.

**Related Jira**
https://issues.jboss.org/browse/RHMAP-9521

**Validation**
Deploy version 4.0.8-93 of the nagios docker image within a Core and MBaaS project. Ensure all expected checks are displayed and passing


mikenairn:
Verified locally, everything working as expected, looks good :+1:

PR #28, opened 2016-08-18
add mysql check

jessesarn:
Also, added additional mongodb container name variation (handles most recent single node mbaases)

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -47,6 +47,11 @@ define command {
   command_line   /opt/rhmap/nagios/plugins/mongodb-health --container $ARG1$
 }
 
+define command {
+  command_name   check_mysql
+  command_line   /opt/rhmap/nagios/plugins/mysql-health --service $ARG1$ --containers $ARG2$ 
+}
+
 define contact {
        contact_name   rhmapadmin
        use            generic-contact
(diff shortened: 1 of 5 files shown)


plugins/default/lib/openshift.py
```diff
@@ -72,7 +72,8 @@ def _get_running_pod_containers(oc, project, selector=None):
     for p in pods:
         if p["status"]["phase"] == "Running":
             for c in p["spec"]["containers"]:
-                result.append((p["metadata"]["name"], c["name"], c))
+                if (not container_names or c["name"] in container_names):
+                    result.append((p["metadata"]["name"], c["name"]))
```
mikenairn: Need to add back returning the container object (c), the memory and cpu checks use it see https://github.com/feedhenry/nagios-container/blob/master/plugins/default/lib/memory_usage.py#L101:

```
result.append((p["metadata"]["name"], c["name"], c))
```

fhservices.cfg.j2
```diff
@@ -139,15 +144,24 @@ define service {
 }
 
 define service {
-       service_description Database::MongoDB::Health
-       check_command check_mongodb!mongodb,mongodb-service
+       service_description mongodb:health
+       check_command check_mongodb!mongodb,mongodb-service,mongodb-1
```
mikenairn: Is this supposed to fix the currently failing health check for mongodb on the core? I'm still seeing an error:

```
CRIT: There are 0 nodes claiming to be primary members of the replica set
```
Created a ticket about this issue also https://issues.jboss.org/browse/RHMAP-9900

mikenairn:
@jessesarn The mysql check is working as expected, but the disk and memory checks are failing because of the issue I commented on.

plugins/default/lib/openshift.py
jessesarn: Ah sorry, I didn't catch that in the rebase that you had added the extra parameter for the container object, will fix during rebase again and verify the cpu/memory checks are good.

fhservices.cfg.j2
jessesarn: Yes, this will fix RHMAP-9900.  Its the 3rd permutation of naming we have between mbaas and core.  I found it only when I used the latest core template again so I just fixed it while I was in there.

PR #31, opened 2016-08-29
RHMAP-9312 - Add container cpu usage nagios check

mikenairn:
* Adds a new nagios check for CPU usage based on the limits set for the container. Uses the time elapsed between checks to work out how much CPU time was used. 
* Will display UNKNOWN status on the initial check since it requires 2 successful checks to take place before it can start calculating usage.

rhmap9312_cpu_checks

Jira:

https://issues.jboss.org/browse/RHMAP-9312

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.pyc
+cpu-usage-data.json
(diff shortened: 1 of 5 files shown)


jessesarn:
LGTM 👍

PR #33, opened 2016-08-30
RHMAP-9901 - Set value used for RHMAP_ROUTER_DNS to 'localhost' if no…

mikenairn:
…ne specified.

Jira:

https://issues.jboss.org/browse/RHMAP-9901


philbrookes:
:+1: Verified locally

PR #34, opened 2016-08-30
added redis check

jessesarn:


jessesarn:
Yes the double $ is intentional and needed to escape nagios trying to replace the value as a variable even though its in single apostrophes.

mikenairn:
Tested locally, everything working as expected :+1: 

redis_check

PR #36, opened 2016-09-01
add mongodb ping check

jessesarn:
JIRA: [RHMAP-9934](https://issues.jboss.org/browse/RHMAP-9934)

Adds mongodb "ping" checks via check_tcp, adds support for make-nagios-fhservices-cfg script to determine the names of the mongodb service(s) based on template parameters (core vs. mbaas).  Defaults to 3 for mbaas if parameter is not found as the 1 node mbaas template doesn't include nagios anyway.

### DIFF
--- a/make-nagios-fhservices-cfg
+++ b/make-nagios-fhservices-cfg
@@ -24,6 +24,13 @@ rhmap_admin_email = os.getenv('RHMAP_ADMIN_EMAIL', 'root@localhost')
 rhmap_router_dns = os.getenv('RHMAP_ROUTER_DNS', 'localhost')
 rhmap_hostgroups = os.getenv('RHMAP_HOSTGROUPS', 'core')
 
+if rhmap_hostgroups == 'mbaas':
+    mongodb_count = int(os.getenv("MONGODB_ENDPOINT_COUNT", 3))
+else:
+    mongodb_count = 1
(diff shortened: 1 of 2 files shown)


make-nagios-fhservices-cfg
```diff
@@ -24,6 +24,13 @@ rhmap_admin_email = os.getenv('RHMAP_ADMIN_EMAIL', 'root@localhost')
 rhmap_router_dns = os.getenv('RHMAP_ROUTER_DNS', 'localhost')
 rhmap_hostgroups = os.getenv('RHMAP_HOSTGROUPS', 'core')
 
+if rhmap_hostgroups == 'mbaas':
+    mongodb_count = int(os.getenv("MONGODB_ENDPOINT_COUNT", 3))
```
mikenairn: MONGODB_ENDPOINT_COUNT isn't being set in any template that we are using, so this will always be 3, is there any point checking this if we are never going to set it?

jessesarn: We are going to set it.  It is in the template variables as ENDPOINT_COUNT, and only gets pulled into the environment variables for mongodb-initator, but assuming we move the mbaas to use configmaps like we do in core, then the nagios pod can automatically have access to that value.

PR #40, opened 2016-09-15
Consistent naming of ping checks

mikenairn:
Rename redis and memcached ping to match all other checks and use uppercase p ping -> Ping


jessesarn:
👍

PR #41, opened 2016-09-15
RHMAP-10225 - Updated host-svc-check helper script

mikenairn:
Updated host-svc-check helper script to take optional args for times and delay. Can be useful for testing certain scenarios to have control over this.

From inside container:
```
$ /opt/rhmap/host-svc-check --times=1 --delay=5
Forcing service checks on 'localhost' (1/1) ...
Waiting 5s ...
Done, all service checks should be up to date.
```

From OS VM:
```
$ oc exec -ti "$(oc get pods | grep nagios | awk '{print $1}')" -- /opt/rhmap/host-svc-check --times=1 --delay=5
```

### DIFF
--- a/scripts/host-svc-check
+++ b/scripts/host-svc-check
@@ -1,23 +1,40 @@
 #!/usr/bin/env python
 import os
 import time
+import argparse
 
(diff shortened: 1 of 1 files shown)


jessesarn:
Looks and works great 👍

PR #42, opened 2016-09-27
added support to find status value from myState or state before treat…

jessesarn:
…ing as unknown

[https://issues.jboss.org/browse/RHMAP-10547]

A mongo node in REMOVED status returns a completely different format object than what the mongodb docs say it should ([https://docs.mongodb.com/manual/reference/command/replSetGetStatus/#rs-status-output]).  Instead of getting the node's state from myState, its called state instead.  The logic now looks for a myState first (since this is what mongo documents it should use), then state (if possible), and only resorting to returning None if neither of those produced a usable value.  This should hopefully cover any other states which don't follow the documented output format.

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -175,7 +175,7 @@ define service {
 
 define service {
        service_description mongodb::replicaset
-       check_command check_mongodb!mongodb,mongodb-service,mongodb-1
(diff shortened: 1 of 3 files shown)


philbrookes:
:+1: Verified on my local machine.

PR #48, opened 2017-03-09
update args to fix error message in mbaas

laurafitzgerald:
**JIRA**
https://issues.jboss.org/browse/RHMAP-14104

@psturc 

Also removing an out of date comment


psturc:
@laurafitzgerald cool, do you think it would be also useful to add another comment that we had to change this param due to changes in `oc` client?
@PhilipGough some more thoughts on this?

PhilipGough:
Yeah a good idea I think In case anything pops up here in future

PhilipGough:
Looks good, we can get this verified and cherry picked over. Thanks @laurafitzgerald

PR #50, opened 2017-03-22
usage of MONGODB_ENDPOINT_COUNT in template

odra:
Uses "MONGODB_ENDPOINT_COUNT" env var in jinja template to include or not replica set monitor (default value is 3).

This wat env var can be set from openshift template param.


PhilipGough:
@odra Good to go 👍

PR #51, opened 2017-04-21
gets memory from /proc/meminfo

odra:
wip


odra:
retest this please

luigizuccarelli:
@odra - One question ? - have you tracked the values you get against the actual heap in the VM. I'm not sure that this will do it - its is still just monitoring the container rather than the vm

odra:
@luigizuccarelli I was able to get the max jvm heap size by running `java -XX:+PrintFlagsFinal -version` (so we can compare this value with the used memory), will update the pr in a few minutes.

PR #57, opened 2017-09-11
Add Jenkinsfile with openshift build image stage.

mikenairn:
Backport of https://github.com/feedhenry/nagios-container/pull/56

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,17 @@
+#!groovy
+
+// https://github.com/feedhenry/fh-pipeline-library
+@Library('fh-pipeline-library') _
+
+final String COMPONENT = "nagios4"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,17 @@
+#!groovy
+
+// https://github.com/feedhenry/fh-pipeline-library
+@Library('fh-pipeline-library') _
+
+final String COMPONENT = "nagios4"
+final String VERSION = "4.0.8"
```
grdryn: Does this version have to be kept in sync with a version anywhere else?

mikenairn: This is the nagios version being installed, if that changes then this will need to be updated.

PR #60, opened 2017-09-27
ios check should use internal dns

PhilipGough:
Nagios should do Jenkins check on sdn as opposed to hitting the external route


odra:
LGTM

PR #66, opened 2018-02-02
remove project check

StephenCoady:
**JIRA**
https://issues.jboss.org/browse/RHMAP-19337

**Changes**
by removing the project check (as it is irrelevant now) the python should always pick up on the env var set from the installer.


cfoskin:
looks good - will verify before giving the ok to merge

PR #75, opened 2018-07-13
Adding new plugin for pod affinity validation

pmccarthy:
**Summary**
In a HA Openshift environment a component with multiple identical pods should have these pods running on separate nodes. This check will issue a warning message should it find two identical pods running on the same node. It is then up to the cluster administrator to determine whether or not these pods should be moved/re-scheduled.

**Validation**
Scale a component to a minimum of 2 pods on the same Openshift node. The expected outcome of the check should be similar to the below example:
```
WARN: redis has 2 pods running on the same node: <node-name>
```
Now schedule these 2 pods to run on separate Openshift nodes. The expected outcome should be:
```
OK: Component pod distribution as expected
```

### DIFF
--- a/plugins/default/lib/openshift.py
+++ b/plugins/default/lib/openshift.py
@@ -7,7 +7,8 @@ def oc(*args):
 
 
 def _get_service_selectors(oc, project, service):
-    svc = json.loads(oc("-n", project, "get", "service", service, "-o", "json"))
+    svc = json.loads(
+        oc("-n", project, "get", "service", service, "-o", "json"))
     return [k + "=" + v for k, v in svc["spec"]["selector"].items()]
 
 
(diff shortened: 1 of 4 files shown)


plugins/default/lib/openshift.py
```diff
@@ -7,7 +7,8 @@ def oc(*args):
 
 
 def _get_service_selectors(oc, project, service):
-    svc = json.loads(oc("-n", project, "get", "service", service, "-o", "json"))
+    svc = json.loads(
```
pmccarthy: Some of these formatting changes are due to running autopep8 against the file as per the recommendations in the README of this repo.

Example:
```
$ autopep8 plugins/default/lib/openshift.py --in-place
```

PR #80, opened 2018-08-28
modified check intervals from 10 to 1/5

jessesarn:
## JIRA
https://issues.jboss.org/browse/RHMAP-20519

## What
Modify check interval of services (1 for database services, 5 for all other)
Don't count the `/etc/hosts` to create disk usage alerts in the Nagios

## Why
The 10 minute interval is too long before NOC may be informed.  Also increase reliability of status information when following status-dependent procedures in documentation and/or automation.

## How
Add new service templates, one for database services which overrides check interval to 1 and another for the other services which overrides check interval to 5.

## Verification Steps
1. Log into nagios
2. Bring up service details and note latest check times
2. Wait at least 60 seconds
3. Refresh the service status details and compare latest check times of mysql and mongodb checks to confirm that the latest check times changed.
4. Wait another 4 minutes
5. Refresh the service status details and compare latest check times of all services to confirm that the latest check times changed

## Checklist:

- [x] Code has been tested locally by PR requester
- [ ] Changes have been successfully verified by another team member 

## Progress

- [] Finished task
- [x] TODO

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -169,11 +169,25 @@ define host {
        hostgroups {{ rhmap_hostgroups }}
 }
 
+# Service templates
+define service {
+       name database-service
+       use generic-service
+       normal_check_interval 1
+       register 0
+}
+define service {
+       name fh-service
+       use generic-service
+       normal_check_interval 5
+       register 0
+}
+
 {% for fhservice in fh_services if 'ping' in fhservice['checks'] %}
 define service {
        service_description {{ fhservice['name'] }}::Ping
        check_command check_fh_component_http_ping!{{ fhservice.get("service", fhservice['name']) }}!{{ fhservice.get("port", "8080") }}!{{ fhservice.get("ping_endpoint", "/sys/info/ping") }}
-       use generic-service
+       use fh-service
        hostgroup_name {{ ",".join(fhservice['hostgroups']) }}
        notes This server cannot access {{ fhservice['name'] }}, check it is running and that this server is able to talk to it on port 8080
        contact_groups rhmapadmins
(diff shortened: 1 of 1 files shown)


fhservices.cfg.j2
```diff
@@ -169,11 +169,25 @@ define host {
        hostgroups {{ rhmap_hostgroups }}
 }
 
+# Service templates
+define service {
+       name database-service
+       use generic-service
+       normal_check_interval 1
```
brizrobbo: Why 1 for this?

jessesarn:
@brizrobbo The reason for the 1 for database services is that its going to improve reliability for at least 2 or 3 other tickets we are trying to target for 4.7.0 release as well as possibly help with 2 customer issues that I'm aware of.

jessesarn:
@brizrobbo I only deployed it locally so far, but I'll go ahead and patch it onto rhm-eng-a so you can have a look there too.

jessesarn:
@brizrobbo I've updated the nagios in rhmap-core on rhm-eng-a to use the dev image.  Feel free to take it for a test drive.

grdryn:
@jessesarn dont' forget that when merging a PR in a component like this, any openshift template PRs that get auto-created need to be manually merged. In this case, there are these two (they got linked above, and you'd have gotten a notification about them when they were created):

https://github.com/feedhenry/fh-openshift-templates/pull/293
https://github.com/fheng/fh-core-openshift-templates/pull/1458

Since you didn't create them, if they need review approval, you can provide that yourself and then merge (and delete the branches). Thanks!

jessesarn:
@grdryn thx, I wasn't sure how these got officially incorporated into master with the new flow compared to how it was prior to the jenkins improvements.  In this case, both of these are superseded by another PR which is using the latest nagios image tag which adds more beyond this.

PR #82, opened 2018-08-31
 ✨ add checks for the mongodb in the user data namespace

wei-lee:
# JIRA

https://issues.jboss.org/browse/RHMAP-21515

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN yum install -y epel-release && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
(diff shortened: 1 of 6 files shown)


wei-lee:
@pmccarthy @jessesarn can you please review this and perhaps test it somewhere? I will try it in the csteam2 cluster and I will expect the checks to fail. Would be good to verify it in a cluster where the checks will pass.

jessesarn:
I'll update the nagios on rhm-eng-a for rhmap-core to use this image version and let you know.

Nanyte25:
I am checking it on rhm-eng-a aswell @jessesarn and @wei-lee

jessesarn:
One other problem while I'm still looking at why the parameter didn't get through.  We need to add userdata or some similar string to the check names.  Right now, the mongodb-<1,2,3>::Ping checks get the same name as the system mongos so there ends up only being 3 checks instead of 6.

On the bright side, the 3 userdata mongo pings overwrote the system ones and are the ones actually getting tested and passing right now.

PR #83, opened 2018-08-31
Rhmap 21515 mongo monitor

jessesarn:
https://issues.jboss.org/browse/RHMAP-21515

Had to resolve a few problems which were preventing mongo checks from working with alternate projects.  One check is still going to fail on OSD but thats due to service account permissions which will need to be addressed as an upgrade/install step.


grdryn:
I see that the templates PRs for this got merged, but this actually didn't, so this isn't actually tagged into 4.7 RC1.

grdryn:
I've merged this into master to keep the master branch consistent with the master branch of the openshift template repos.

jessesarn:
Sorry for the oversight.  Thx @grdryn.

PR #84, opened 2018-09-05
mysql service list

odra:
Uses a list o generate mysql nagios checks for all hosts

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -338,29 +338,16 @@ define service {
 }
 {% endfor %}
 
-{% if mysql_services|length == 1 %}
+{% for mysql_service in mysql_services %}
 define service {
-       service_description mysql::health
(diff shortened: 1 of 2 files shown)


odra:
@jessesarn I noticed we are always adding the userspace mongodb checks: https://github.com/feedhenry/nagios-container/blob/4a49e091812b34084e90db3787a42dbb533977a1/make-nagios-fhservices-cfg#L41 - should we add some kind of check to not include those if the running rhmap instance did not deploy the userpace parts?

jessesarn:
@odra Yes, I think we should avoid adding hosted-specific checks to self-managed installs.

odra:
Ok, I will create another jira ticket to fix the userspace mongodb check to keep the pr clean

jessesarn:
@odra cool thx

PR #88, opened 2018-09-27
👽 Remove --show-all for oc 3.10

grdryn:
JIRA: https://issues.jboss.org/browse/RHMAP-21670

In `oc` 3.10, if the `--show-all` parameter is used for `oc get`, then a deprecation warning is output along with the regular output. When this python script tries to decode the whole output to JSON, it fails (even though the regular output without the warning is valid and decode-able).

Given that this parameter appears to have been arbitrarily changed from `--show-all=false` to `--show-all=true` in [1] just to get around some other issue with a different `oc` version, I think it's fine to just remove it.

[1] https://github.com/feedhenry/nagios-container/pull/48


camilamacedo86:
Testing the image : 4.0.8-cf0a6e2

PR #90, opened 2018-09-27
Rhmap 21696 mysql check fix

odra:
Fixes mysql checks for SM deployments.

Verification:

* Deploy the generated container image against a SM RHMAP instance - only one mysql check should be listed
* rsh into the nagios pod and check the services file: /etc/nagios/conf.d/fhservices.cfg (only one mysql service check should be defined there)

### DIFF
--- a/fhservices.cfg.j2
+++ b/fhservices.cfg.j2
@@ -338,6 +338,18 @@ define service {
 }
 {% endfor %}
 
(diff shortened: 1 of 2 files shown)


camilamacedo86:
I am testing it. Could you just please add the JIRA for this issue here, and if do not exist create one. It is important for us to track what was done.

PR #92, opened 2018-10-04
RHMAP-20790 mongodb volume check

davidkirwan:
JIRA: https://issues.jboss.org/browse/RHMAP-20790
JIRA: https://issues.jboss.org/browse/RHMAP-20595

### DIFF
--- a/plugins/default/lib/disk_usage_mongo.py
+++ b/plugins/default/lib/disk_usage_mongo.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python
+import argparse
+import re
+import sys
+import traceback
+from collections import Counter
+
+import nagios
+import openshift
+
+
+def generate_parser():
+    parser = argparse.ArgumentParser(
+        description="Checks the disk usage (blocks and inodes)",
+    )
+    parser.add_argument(
+        "-w", "--warn", type=int, required=True,
+        help="set warning threshold of disk usage (%% of blocks or inodes)",
+    )
+    parser.add_argument(
+        "-c", "--crit", type=int, required=True,
+        help="set critical threshold of disk usage (%% of blocks or inodes), "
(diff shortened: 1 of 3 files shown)


plugins/default/lib/disk_usage_mongo.py
```diff
@@ -0,0 +1,133 @@
...
+    if mo is None:
+        return ()
+    return mo.group(3), int(mo.group(1)), int(mo.group(2))
+
+
+def parse_df_lines(lines):
+    return filter(None, map(parse_df_line, lines.splitlines()))
+
+
+def analize(pod, container, disks, warning_threshold, critical_threshold):
```
damienomurchu: `analyze` ?

davidkirwan: Copy paste from another similar check :face_with_head_bandage:

damienomurchu: If the code is identical/ very similar - maybe it could re-used ? (unless there's a time-sensitive element to these changes of course)

davidkirwan: I think without some serious refactoring this is the quickest way to handle this particular requirement of targeting the particular mongo nodes. I'd like to get this into the 4.7.1 release.

grdryn: @davidkirwan I see that one of the linked issues has a fix version of 4.X-Future, and the other one doesn't have a fix version. I know @davidffrench did a lot of work grooming the backlog yesterday, so I'd suggest checking with him before merging anything into the FH-v4.7 branch for a 4.7.1.

davidffrench: This one might have already been in progress before I got to it. Changed the fixversion to 4.7-Future. I don't see any reason not to include it in a patch release.

